PR #12, opened 2021-09-10
Issue 1 gene

korikuzma:


server/app/graphql/types/gene_alias_type.rb
```diff
@@ -0,0 +1,7 @@
+module Types
+    class GeneAliasType < Types::BaseObject
```
nairod2000, line 2: The convention for spacing has been to do two spaces instead of four

server/app/graphql/types/gene_attribute_type.rb
```diff
@@ -0,0 +1,8 @@
+module Types
+    class GeneAttributeType < Types::BaseObject
+        field :id, ID, null: false
+        field :gene_id, String, null: false
```
nairod2000, line 4: We should maybe represent the foreign key fields as being of type ID instead of String just to be consistent

nairod2000 (approved):
The types look good! Moving forward with them, their relationships to other graphql types will need to be added to their fields

PR #13, opened 2021-09-13
Issue 1 gene

korikuzma:
Address PR review suggestions

### DIFF
--- a/server/app/graphql/types/gene_gene_interaction_claim_type.rb
+++ b/server/app/graphql/types/gene_gene_interaction_claim_type.rb
@@ -8,6 +8,6 @@ module Types
     field :source, Types::SourceType, null: false
     field :interacting_gene, Types::GeneType, null: false
     field :gene, Types::GeneType, null: false
-    field :gene_gene_interaction_claim_attributes, [Types::GeneGeneInteractionClaimAttributeType]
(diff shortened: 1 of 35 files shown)


server/app/graphql/types/gene_gene_interaction_claim_type.rb
```diff
@@ -8,6 +8,6 @@ class GeneGeneInteractionClaimType < Types::BaseObject
     field :source, Types::SourceType, null: false
     field :interacting_gene, Types::GeneType, null: false
     field :gene, Types::GeneType, null: false
-    field :gene_gene_interaction_claim_attributes, [Types::GeneGeneInteractionClaimAttributeType]
+    field :gene_gene_interaction_claim_attributes, [Types::GeneGeneInteractionClaimAttributeType], null: true
```
susannasiebert, line 11: For is-many fields, we've found that it's better to set them to `null:false`, even if they're technically nullable. If there are no `gene_gene_interaction_claim_attributes` this would then return an empty array instead of `null`, which makes things more straightforward on the frontend side.

korikuzma: Sounds good. Will make this change.

nairod2000 (approved):
Looks good!

PR #17, opened 2021-09-27
Issue 1

nairod2000:
Moved over the models and made graphql objects for them


acoffman (approved):
Probably want one more pass through to add missing loaders for a few entities, and maybe remove some unused code, but on the whole this looks like a good start to me! I don't see any reason not to merge.

PR #85, opened 2022-04-11
Feat: Catch invalid claim types/categories with logging, not exceptions

### LINKED ISSUE #79: Figure out/set up logging
Knowing Rails, this is probably both a) configured out of the box and b) already in use somewhere, but we should get it worked out and put in place (eg for handling invalid claim types)

jsstevenson:
Close #79 
* Adds some extra logging for the thing I was thinking of (how invalid interaction claim types/gene categories are caught)

### DIFF
--- a/server/lib/genome/importers/base.rb
+++ b/server/lib/genome/importers/base.rb
@@ -63,9 +63,15 @@ module Genome
       def create_gene_claim_category(gene_claim, category)
         gene_category = GeneClaimCategory.find_by(name: category)
         if gene_category.nil?
-          raise StandardError, "GeneClaimCategory with name #{category} does not exist. If this is a valid category, please create its database entry manually before running the importer."
+          msg = "Unrecognized GeneClaimCategory #{category} from #{gene_claim.inspect}."
+          raise StandardError, msg unless Rails.env == 'development'
+
+          puts msg
+          Rails.logger.debug msg
         else
-          gene_claim.gene_claim_categories << gene_category unless gene_claim.gene_claim_categories.include? gene_category
+          unless gene_claim.gene_claim_categories.include? gene_category
+            gene_claim.gene_claim_categories << gene_category
+          end
         end
       end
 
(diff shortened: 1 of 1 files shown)


server/lib/genome/importers/base.rb
```diff
@@ -63,9 +63,15 @@ def create_gene_claim_attribute(gene_claim, name, value)
       def create_gene_claim_category(gene_claim, category)
         gene_category = GeneClaimCategory.find_by(name: category)
         if gene_category.nil?
-          raise StandardError, "GeneClaimCategory with name #{category} does not exist. If this is a valid category, please create its database entry manually before running the importer."
+          msg = "Unrecognized GeneClaimCategory #{category} from #{gene_claim.inspect}."
+          raise StandardError, msg unless Rails.env == 'development'
+
+          puts msg
+          Rails.logger.debug msg
```
mcannon068nw, line 70: puts!

server/lib/genome/importers/base.rb
```diff
@@ -110,11 +116,15 @@ def create_interaction_claim_type(interaction_claim, type)
...
+          raise StandardError, msg unless Rails.env == 'development'
 
-        unless interaction_claim.interaction_claim_types.include? claim_type
-          interaction_claim.interaction_claim_types << claim_type
+          puts msg
+          Rails.logger.debug msg
+        else
+          unless interaction_claim.interaction_claim_types.include? claim_type
+            interaction_claim.interaction_claim_types << claim_type
+          end
```
mcannon068nw, line 127: puts?

or are these puts for error handling

mcannon068nw (approved):

server/lib/genome/importers/base.rb
```diff
@@ -110,11 +116,15 @@ def create_interaction_claim_type(interaction_claim, type)
           type: Genome::Normalizers::InteractionClaimType.name_normalizer(type)
         )
         if claim_type.nil?
-          raise StandardError, "InteractionClaimType with type #{type} does not exist. If this is a valid type, please create its database entry manually before running the importer."
-        end
+          msg = "Unrecognized InteractionClaimType #{type} from #{interaction_claim.inspect}"
+          raise StandardError, msg unless Rails.env == 'development'
```
acoffman, line 120: This will work just fine as written, however you can also just do `Rails.env.development?` to get a boolean directly, rather than manually doing an equality check

PR #110, opened 2022-05-06
fix: correct directionality enum mapping for graphql

### LINKED ISSUE #97: Directionality returns as "0" for both inhibiting and activating
Directly querying the db in rails console returns directionality `"activator"`, `"inhibitory"`,  or `"nil"`. The corresponding enum values expected from graphql are `0` for activator, `1` for inhibitory,  and `null` for nil. 

While 0 is returned for activator types (and null for nil types), graphql seems to also return 0 for inhibitory types.

jsstevenson:
close #97 

I think this is the correct way to do this (@acoffman)?

Also, it looks to me like "directionality unclear" isn't used by any importers, maybe delete?

### DIFF
--- a/server/app/graphql/types/directionality_type.rb
+++ b/server/app/graphql/types/directionality_type.rb
@@ -1,0 +1,6 @@
+module Types
+  class DirectionalityType < Types::BaseEnum
+    value :activating
+    value :inhibitory
+  end
+end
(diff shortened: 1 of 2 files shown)


acoffman (dismissed):
This looks broadly correct to me. The only thing I would say is that GraphQL enums are generally all caps. If you want to map the GraphQL representation to a different value in ruby, you can provide a value argument. Example here: https://github.com/griffithlab/civic-v2/blob/main/server/app/graphql/types/events/event_action_type.rb

That way in the GraphQL schema it will be `ACTIVATING`, but anywhere in ruby it will be `activating` like you expect.

acoffman (approved):

PR #111, opened 2022-05-06
fix: call drug claim alias creator correctly for pharmgkb import

### LINKED ISSUE #103: Fix PharmGKB aliases
The PharmGKB importer calls `create_drug_claim_alias` w/ the wrong order of arguments, just need to swap them

jsstevenson:
close #103


mcannon068nw (approved):
looks good! simple elegant fixes

PR #122, opened 2022-05-11
Precompute Interaction Scores

acoffman:
In the common case (no partners are passed in) we can calculate the interaction score at grouping time (or on demand) and return it directly rather than on the fly.


mcannon068nw (approved):
looks good. Per Adam's earlier message, we should consider looking into further optimizing this to speed up run time

jsstevenson (approved):
👍 

FWIW, running a clean interaction grouping run now takes about 90 minutes total now (couldn't say what it was before but that's not a totally intolerable benchmark)

PR #124, opened 2022-05-12
ci: add workflow helpers

jsstevenson:
* add test that PRs into `main` come from `staging`, and that PRs into `staging` come from `dev`
* add action that closes related issues when PRs are merged to `dev`


mcannon068nw (approved):
let's try it out!

PR #125, opened 2022-05-12
Initial Bootstrap of Rspec/factory_bot

acoffman:
This removes the unit test and fixtures stubs that were auto generated and installs `rspec-rails` as well as `factory_bot`. It also provides some helper config for hooking it up together and testing graphql.

Tests can be run with `rails spec`. I've included a (very) simple example of an end-to-end test. It consists of a factory for a gene, then a test that generates a gene using that factory and queries the `genes` graphql endpoint via its name.

Documentation on how to make your own factories can be found here: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#defining-factories

and some handy  documentation on the types of assertions you can make can be found here: https://relishapp.com/rspec/rspec-expectations/docs

It should also be noted that the `before(:example)` is run before each test (example), and the whole thing is wrapped in a transaction that will be rolled back at the end of the example. In that way each `it` block starts with the database in exactly the same state regardless of what the previous test may have done.

I'm more than happy to provide additional examples or answer any questions you guys might have!


mcannon068nw (approved):
looks great! Thank you for setting this up Adam!

PR #129, opened 2022-05-16
Issue 88

cjosu:
Building out drug results page, more or less mirrors gene results for the time being.


jsstevenson (changes requested):
Looks like the filetree issue snuck in here: 

https://github.com/dgidb/dgidb-v5/tree/issue-88/client/src/pages

I was able to resolve this in another branch by nuking the entire directory, committing/pushing that, and then manually recreating the files with the correct directory structure and casing, then recommitting them

PR #138, opened 2022-05-18
feat: add example query to graphiql

jsstevenson:


mcannon068nw (approved):
example query looks great

PR #143, opened 2022-05-20
Drop primary name from drug claim

jsstevenson:
`primary_name` is equal to `name` for all drug claims other than normalizer-provided sources, and GuideToPharmacology. This seems unnecessary.


mcannon068nw (approved):
I think this makes sense to change. When making tweaks to the importers I always found it weird entering the drug name twice in the args

PR #158, opened 2022-05-23
feat: add remaining normalizers

### LINKED ISSUE #90: Add post-group routines: source counters, normalize drug types, source trust levels, interaction types

jsstevenson:
Close #90 

 * Get source counts and drug type normalizers up and running
    * Update drug type normalization to include a bunch of additional values. There's more room for refinement here, and/or normalization at the import level
    * Added a couple of additional antineoplastic sources -- this may or may not be a good idea
 * Remove trust levels, interaction claim types, and publications normalizers -- they were already being run on a per-source basis within each importer (I'm a little confused by publications so we can circle back to that later)
 * Remove drug claim type and gene-gene interaction tables. As far as we could tell, they weren't getting used anywhere.


server/lib/genome/normalizers/drug_type.rb
```diff
@@ -17,58 +16,73 @@ def self.normalize_types
...
+          'CIViC',
           'ClearityFoundationBiomarkers',
           'ClearityFoundationClinicalTrial',
-          'CKB',
+          'DoCM',
+          'MyCancerGenome',
+          'MyCancerGenomeClinicalTrial',
           'OncoKB',
-          'CGI'
+          'TALC'
```
jsstevenson, line 85: I rearranged the order so it looks like there are more changes, but the new sources here would be DoCM and CIViC

server/lib/genome/normalizers/interaction_claim_type.rb
```diff
@@ -71,42 +71,52 @@ def self.name_normalizer(val)
...
           'negative modulator'
-        when 'inhibitory immune response', 'car-t-cell-therapy(dual specific)', 'immunomodulator', 'immunomodulator (immunostimulant)', 'immune response agent', 'car-t-cell-therapy', 'immune response agent', 'immunostimulant', 'immunostimulator', 'Radioimmunotherapy'
+        when 'inhibitory immune response', 'car-t-cell-therapy(dual specific)', 'immunomodulator',
+            'immunomodulator (immunostimulant)', 'immune response agent', 'car-t-cell-therapy',
+            'immunostimulant', 'immunostimulator', 'Radioimmunotherapy'
           'immunotherapy'
         when 'component of'
           'product of'
-        when 'opener'
+        when 'opener'  # TODO: duplicate w/ 'other/unknown' -- how to resolve?
```
jsstevenson, line 106: 'opener' is included with the 'other/unknown' group above, so this can't be reached -- we should remove it from one or the other groupings

mcannon068nw (approved):

PR #167, opened 2022-06-06
Issue 127

cjosu:
Inconsistencies between dev's structure and my local may cause some issues


mcannon068nw:
everything builds and runs for me too

PR #171, opened 2022-06-08
feat: use CIViC v2 API for imports

### LINKED ISSUE #130: CIViC importer/updater use v2 endpoint
Old endpoint won't include any new data moving forward

jsstevenson:
close #130


mcannon068nw (approved):
looks good.

tried running it and did not get any weird timeouts or anything like we had with a certain other importer

PR #181, opened 2022-06-19
chore: fix alias namespaces

jsstevenson:
eg EWSR1 has alias "2130" https://dgidb.org/genes/EWSR1 -> should be "ncbigene:2130". This also slightly increases the normalizers' capture rates.

### DIFF
--- a/server/lib/genome/groupers/drug_grouper.rb
+++ b/server/lib/genome/groupers/drug_grouper.rb
@@ -137,12 +137,32 @@ module Genome
         response['therapy_descriptor']['therapy_id'] unless response['match_type'].zero?
       end
 
+      def produce_concept_id_nomenclature(concept_id)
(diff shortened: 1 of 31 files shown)


server/lib/genome/groupers/drug_grouper.rb
```diff
@@ -137,12 +137,32 @@ def get_concept_id(response)
...
+        when /drugsatfda/
+          'Drugs@FDA ID'
+        when /chemidplus/
+          'ChemIDplus ID'
+        when /wikidata/
+          'Wikidata Drug ID'
+        else
+          'Concept ID'
+        end
+      end
```
mcannon068nw, line 157: random but I the ruby 'case when' setup is so satisfying

mcannon068nw (approved):
looks good

Issue #105, opened 2022-05-02
Standardize case


jsstevenson:
Some table columns are all-caps, others aren't. We should make sure we understand why and adjust accordingly.

Issue #104, opened 2022-04-30
Gene aliases


jsstevenson:
There are a lot of them and they may not all be meaningful -- either including nomenclature, or limiting what kinds of aliases get lifted up to the Gene groups might help reduce the infodump on the client's perspective

Issue #65, opened 2022-04-04
Clean up gene and drug aliases


jsstevenson:
In particular, try to identify cases where non-namespaced ID numbers are getting grouped into genes and drugs and fix the importer code accordingly

Issue #44, opened 2022-03-23
Process remaining new interaction claim types and gene categories


jsstevenson:
**InteractionClaimType** -- normalization defined in interaction claim type model
Clarity Biomarkers: "Biomarker"
Clarity Clinical Trials: "immunostimulator", "natigen", "radioimmunotherapy"
My Cancer Genome: "immunotherapy"

**GeneCategory**
Hopkins/Groom: "DNA DIRECTED DNA POLYMERASE"

PR #215, opened 2022-11-04
fix: data cleaning

### LINKED ISSUE #210: Data cleaning

jsstevenson:
Close #210, #104, #105, #65, #44, #27

A few outstanding questions for the neighborhood pharmacology expert but should be smooth otherwise. I also flagged one design question for Adam.

### DIFF
--- a/server/db/seeds.rb
+++ b/server/db/seeds.rb
@@ -34,22 +34,16 @@ interaction_claim_types = [
     reference: 'DrugBank examples: <a href="http://www.ncbi.nlm.nih.gov/pubmed/12070353">pubmed:12070353</a>'
   },
   {
-    type: 'n/a',
-    directionality: nil,
-    definition: 'DGIdb assigns this label to any drug-gene interaction for which the interaction type is not specified by the reporting source.',
-    reference: nil
-  },
-  {
     type: 'negative modulator',
     directionality: 1,
     definition: 'In a negative modulator interaction, the drug negatively regulates the amount or activity of its target. In contrast to an inhibitory allosteric modulator, this interaction type may not involve any direct binding to the target.',
     reference: '<a href="https://en.wikipedia.org/wiki/Allosteric_modulator">Wikipedia - Allosteric Modulator</a>'
   },
   {
-    type: 'partial agonist',
-    directionality: 0,
-    definition: 'In a partial agonist interaction, a drug will elicit a reduced amplitude functional response at its target receptor, as compared to the response elicited by a full agonist.',
-    reference: '<a href="https://en.wikipedia.org/wiki/Receptor_antagonist#Partial_agonists">Wikipedia - Receptor Antagonist</a>'
+    type: 'antibody',
+    directionality: 1,
+    definition: 'TODO',
+    reference: 'TODO'
   },
   {
     type: 'adduct',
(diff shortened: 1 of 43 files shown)


server/db/seeds.rb
```diff
@@ -33,23 +33,17 @@
...
   },
   {
-    type: 'partial agonist',
-    directionality: 0,
-    definition: 'In a partial agonist interaction, a drug will elicit a reduced amplitude functional response at its target receptor, as compared to the response elicited by a full agonist.',
-    reference: '<a href="https://en.wikipedia.org/wiki/Receptor_antagonist#Partial_agonists">Wikipedia - Receptor Antagonist</a>'
+    type: 'antibody',
+    directionality: 1,
+    definition: 'TODO',
+    reference: 'TODO'
```
jsstevenson, line 46: @mcannon068nw input pls

server/db/seeds.rb
```diff
@@ -200,6 +188,24 @@
...
+    type: 'immunotherapy',
+    directionality: nil,
+    definition: 'Immunotherapy is used to upregulate or downregulate the immune system to achieve a therapeutic effect in immunological mediated disorders, including immunodeficiencies, hypersensitivity reactions, autoimmune diseases, tissue and organ transplantations, malignancies, inflammatory disorders, infectious diseases, and any other disease, where immunotherapy can improve the quality and life expectancy.',
+    reference: 'https://www.ncbi.nlm.nih.gov/books/NBK519046/'
+  },
+  {
+    type: 'cytotoxic',
+    directionality: nil,  # TODO
+    definition: 'TODO',  # TODO
+    reference: 'TODO'  # TODO
```
jsstevenson, line 208: @mcannon068nw input pls

server/db/seeds.rb
```diff
@@ -200,6 +188,24 @@
     directionality: nil,
     definition: 'Substrates are drugs which are metabolized by the protein they are interacting with.',
     reference: '<a href="https://www.fda.gov/drugs/drug-interactions-labeling/drug-development-and-drug-interactions-table-substrates-inhibitors-and-inducers">FDA - Table of Substrates, Inhibitors and Inducers</a>'
+  },
+  {
+    type: 'vaccine',
+    directionality: 0,
+    definition: 'In vaccine interactions, the drugs stimulate or restore an immune response to their target.',
+    reference: '<a href="https://www.cancer.gov/about-cancer/treatment/types/immunotherapy#q3">NCI - Cancer Vaccines</a>'
```
jsstevenson, line 196: Would like to revisit this style of data presentation when we cover the Sources page (in I think #170) because it's pretty janky to dump raw html into the database

server/lib/genome/names.rb
```diff
@@ -0,0 +1,120 @@
+module Genome
```
jsstevenson, line 1: @acoffman this seemed like a straightforward way to get something like a shared set of enums (rather than ruby symbols) to standardize values across importers/groupers/normalizers. Is this sufficiently idiomatic? Any suggestions?

server/db/seeds.rb
mcannon068nw: type: 'antibody',
    directionality: 1,
    definition: 'An antibody interaction occurs when an antibody drug specifically binds the target molecule. Most interactions of this class fall under various natural antibody functions: neutralization, antibody-dependent cell-mediated cytotoxicity (ADCC), complement-dependent cytotoxicity (CDC), or drug delivery carriers.', 
    reference: '<a href="https://www.ncbi.nlm.nih.gov/pubmed/26441475/">pubmed:26441475</a>'

server/lib/genome/names.rb
acoffman: You can add 
`# frozen_string_literal: true` 

As the first line in the file rather than `.freeze` everywhere for this sort of thing if you prefer. You _could_ potentially make these dictionaries/hashes, but based on the usage elsewhere in this PR I think I agree that consts are actually better here!

PR #223, opened 2022-11-30
Fixed DrugBank Importer Fetching Blank Values

### LINKED ISSUE #142: Fix DrugBank blank record issue
The DrugBank importer seems to be creating a record with a blank name:

```
SELECT * FROM drug_claims dc
WHERE s.source_db_name = 'DrugBank' AND dc.name = '';
```

That record has a bunch of different drugbank IDs as aliases

```
SELECT dca.alias FROM drug_claims dc
LEFT JOIN sources s on dc.source_id = s.id
         RIGHT JOIN drug_claim_aliases dca on dc.id = dca.drug_claim_id
WHERE s.source_db_name = 'DrugBank' AND dc.name = '';
```

Two steps to take
1) prevent ourselves from adding drug claims with blank names
2) figure out where the label field is for these drug claims so that we can disaggregate them

nairod2000:
closes #142 
I have fixed the problem of the drugbank importer creating DrugClaims and GeneClaims with blank names and InteractionClaimAttributes with blank values.

The blank name for DrugClaims was a result of a flag in the xml parser incorrectly getting flipped to true. The other two cases were due to not having checks for when their tags of interest were blank. 

This PR adds a validation to the name field on DrugClaim, and changes `first_or_create` to `first_or_create!` in /lib/genome/importers/base.rb for the `create_drug_claim` method. We might also want to consider adding validations to the other fields affected by this issue. If that's desired, I can updated this PR or create a new issue and add it there.

### DIFF
--- a/server/app/models/drug_claim.rb
+++ b/server/app/models/drug_claim.rb
@@ -9,6 +9,8 @@ class DrugClaim < ::ActiveRecord::Base
   has_many :drug_claim_attributes, inverse_of: :drug_claim, dependent: :delete_all
   has_many :drug_claim_approval_ratings, inverse_of: :drug_claim, dependent: :delete_all
 
+  validates :name, presence: true, allow_blank: false
+
   def self.for_search
     eager_load(
       drug: [
--- a/server/lib/genome/importers/base.rb
+++ b/server/lib/genome/importers/base.rb
@@ -108,7 +108,7 @@ module Genome
           name: name.strip,
           nomenclature: nomenclature.strip,
           source_id: source.id
-        ).first_or_create
+        ).first_or_create!
       end
 
       def create_drug_claim_alias(drug_claim, synonym, nomenclature)
--- a/server/lib/genome/importers/file_importers/drugbank.rb
+++ b/server/lib/genome/importers/file_importers/drugbank.rb
@@ -120,7 +120,9 @@ module Genome; module Importers; module FileImporters; module Drugbank;
         when 'drugbank-id'
             if attrs['primary']
                 @drugbank_id_flag = true
-                @is_primary_drug = true
+                if @is_not_salts
+                    @is_primary_drug = true
+                end
             end
 
         when 'name'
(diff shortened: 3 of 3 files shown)


jsstevenson:
If possible, I think we want to somehow rescue the records that were coming up with blank names previously -- I did a cursory check and drugs like "glucagon" (https://go.drugbank.com/drugs/DB00040) and "insulin" (https://go.drugbank.com/drugs/DB00030) don't seem to be coming through. Is the raw XML just missing the name for these records, or is it dumping it under a different field or something?

nairod2000:
After running
`i = Genome::Importers::FileImporters::Drugbank::Importer.new('/home/nairod/Downloads/drugbank-full.xml')`
`i.import`
in the rails console these are the queries and outputs I got
Glucagon_and_Human_insulin

Then the importer uses the records with `nomenclature: "DrugBank Drug Name"` for the creation of the interaction claim for each of the genes targeted by that drug. I think these are the records that we are wanting created out of the file.

nairod2000:
Another note: 
The importer extracts about 19,000 drug-gene interactions from the file and before the fix about 5000 of them had a blank name for the drug involved. All of the drugs involved in those interactions got condensed into one drug_claim. Similar with the drug-gene interactions that had a blank gene name, but I did not get a number on how many had that problem. So, there's a lot of interaction claims that should belong to unique drug_claims or gene_claims that got messed up by how the importer was essentially losing the name data and related them all to the one gene and or drug claim that has a blank name. 

To get the correct data in the database I think the drug_claim with `name: ""`, the gene_claim with `name: ""` and interaction claims that are related one or both of those should be deleted, then the importer should be DrugBank importer should be re-run.

jsstevenson:
@nairod2000 my bad, I wasn't casing them correctly. Looks good on my end. I'm going to do a few more spot checks vs the v4 data in the morning but I suspect that we're good to go.

jsstevenson:
👍 looks good. I'll get an updated data dump out by the end of the week.

PR #243, opened 2023-01-06
Sources page rework

katiestahl:
Here is a PR for some enhancements to the Browse Sources page. 
* Fixed styling issues
* Added filtering
* Added page info/heading at the top
* Added License info
* Added Full Citation info
* Fixed/adding missing gene/drug/interaction counts


client/src/components/Browse/Sources/BrowseSources.tsx
```diff
@@ -22,107 +31,128 @@ export const BrowseSources = () => {
...
+
+  const handleButtonClick = (event: any) => {
+    const value = event.target.value
+    setFilter(value)
+  }
 
-  useEffect(() => {
-    console.log('drugData', drugData)
-  }, [drugData])
+  const getCard = (src: any) => {
```
katiestahl, line 62: Same issue here with the "any" type. This one should play nicer than the above event type since this should just be an object. Is the correct practice to define this entire object as the type here? It'll be pretty lengthy

mcannon068nw (commented):
ran locally, everything worked and I think everything looks great (way better than how it previously looked). I will defer to @acoffman or @jsstevenson for your specific comments on proper type usage before approval though

client/src/components/Browse/Sources/BrowseSources.tsx
```diff
@@ -22,107 +31,128 @@ export const BrowseSources = () => {
...
+            variant={filter === "Potentially Druggable" ? "outlined" : "contained"} 
+            value="Potentially Druggable">
+            Potentially Druggable
+          </Button>
+        </ButtonGroup>
+      </Box>
+      {
+        sectionsMap.map((section: any) => {
+          return section.value === filter || filter === "All" ? (
+            <>
```
jsstevenson, line 143: I don't think it matters but I'm a stickler for warnings - this `map` call, and the one on line 147, both throw the react `each child needs a key prop` warning. Should be a pretty easy fix

acoffman (approved):
It looks good to me as well. DOM events in React don't always play nicely with types unfortunately, there's some explanation of the specic problem you're encountering with Material UI here: https://mui.com/material-ui/guides/typescript/#handling-value-and-event-handlers with some further explanation in the linked github repo. 

Its not ideal, but something like this will make the compiler happy and give you a little more safety than `any`: 
Screenshot 2023-01-06 at 1 17 14 PM

client/src/components/Browse/Sources/BrowseSources.tsx
acoffman: As you point out this one is a distinct problem from the event handler, it can be typed. But as you also note, defining the entire type inline here would be cumbersome and annoying. 

This goes back to the link in one of the issues about `any` types where you can generate type definitions for your graphql queries. That way the data you get back in the very beginning from the various`useSources` hooks is already typed (instead of the `any` you currently get back).

At that point you shouldn't really even have to specify the types manually at all, the compiler will infer them and that will propagate throughout the component.

jsstevenson (changes requested):

acoffman:
For more type safety in the future you could imagine doing something like this:

Screenshot 2023-01-06 at 1 41 35 PM

Now the filter is constrained to only those literal values, and the individual sections are typed, likewise the `useState` hook can be constrained to only take valid values rather than any string. It also helps with typos. For instance if i were to forget and type 'Genes' that's now a compile time error:

Screenshot 2023-01-06 at 1 44 35 PM

Without the typed queries, you still have that `sources: any[]` so its probably not worth pursuing until/if we decide to go for typed queries, but just an idea of the kinds of things you can do.

jsstevenson (approved):

PR #245, opened 2023-01-10
fix: handle civic api update

### LINKED ISSUE #244: Ensure compatibility w/ latest CIViC release

jsstevenson:
close #244

### DIFF
--- a/server/lib/genome/importers/api_importers/civic/api_client.rb
+++ b/server/lib/genome/importers/api_importers/civic/api_client.rb
@@ -42,26 +42,30 @@ module Genome; module Importers; module ApiImporters; module Civic
                 name
                 geneAliases
               }
-              evidenceItems {
+              molecularProfiles {
                 nodes {
-                  name
-                  id
-                  clinicalSignificance
-                  evidenceType
-                  evidenceLevel
-                  evidenceDirection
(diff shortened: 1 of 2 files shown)


jsstevenson:
@acoffman is there some kind of guard clause that we can wrap the calls to the client parser with to protect against/emit a warning for future API changes?

acoffman (approved):
This looks good to me overall/should get things working again.

The one thing that might be worth discussion in a meeting is what to do with evidence items that span multiple genes.

Molecular Profiles allow for arbitrary logical combination of Variants. So you could have evidence attached to something like

`Gene 1 Variant 1 AND (Gene 2 Variant 2 OR Gene 3 Variant 3)`

How do we want to create gene claims in this case? Attach to all three? MPs also support `NOT` so if we have:

`NOT Gene 1 Variant 1  AND Gene  2 Variant 2` do we attach to Gene 1 and Gene 2 and denote the `NOT` somehow?

acoffman:
> @acoffman is there some kind of guard clause that we can wrap the calls to the client parser with to protect against/emit a warning for future API changes?

This is a good question. Not currently no, but we _should_ have something like that. I will look into adding a status/metadata query for civic that we can use for that here.

PR #248, opened 2023-01-17
test: add graphql query coverage

jsstevenson:
Still WIP -- ironing a few things out


server/app/graphql/resolvers/genes.rb
```diff
@@ -9,7 +9,7 @@ class Resolvers::Genes < GraphQL::Schema::Resolver
   scope { Gene.all }
 
   option(:ids, type: [String]) { |scope, value| scope.where(id: value)}
-  option(:name, type: String) { |scope, value| scope.where("name ILIKE ?", "#{value}%")}
+  option(:names, type: [String]) { |scope, value| scope.where(name: value) }
```
jsstevenson, line 12: @acoffman is this the correct activerecord call for matching against an array of names? (Also, does anyone have any theories why we were doing an SQL startswith query here?)

server/spec/queries/drug_interaction_query_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'rails_helper'
+
+RSpec.describe 'Drug Interaction Query', type: :graphql do
+  before(:example) do
+    @drug = create(:drug, approved: true)
```
jsstevenson, line 5: @acoffman am I missing something here? the `approved` value refuses to be set (the check on line 54 is failing)

acoffman: I would expect that to work yeah, are there any errors thrown? Does the drug report as valid otherwise?

jsstevenson: Nope, it just... silently retains the `false` value set by the `before_create` hook. Every assertion passes except for the truthiness one on line 55ish.

server/app/controllers/graphql_controller.rb
```diff
@@ -1,4 +1,4 @@
-class GraphQLController < ApplicationController
+class GraphqlController < ApplicationController
   # If accessing from outside this domain, nullify the session
   # This allows for outside API access while preventing CSRF attacks,
   # but you'll have to authenticate your user separately
```
mcannon068nw, line 4: definitely interested into why this is causing us problems across sites haha

mcannon068nw (approved):

PR #250, opened 2023-01-17
fix: resolve oncokb comparison warning

jsstevenson:


mcannon068nw (approved):
This was a bit of a doozy. Going to approve for now but might need to circle back around later to double check @jsstevenson

PR #252, opened 2023-01-18
Record pages rework

katiestahl:


jsstevenson:
Personally I'm ready to approve but I pulled up the v4 versions of these pages and I wanted to flag some stuff from there that we could either a) drop for v5, b) try to get done now, or c) make separate issues for

for genes, I think it's be good to put the concept ID up in the header next to the gene name (maybe greyed out/unbolded, and either spaced away or flexed to the right side). This is not actually a field correctly exposed by graphql right now (lol) so I will make a ticket in a sec

mcannon068nw:
I agree that it would be good to provide the concept id next to the name of the drug at the top (in grey text or some kind of subtext format for sure). The improvements to normalization are intended to be one of the update focuses for v5 so providing that information in a visible location seems prudent.

For the big category pills, those are present and represented in the categories box on gene record pages, but they definitely don't have the same 'pop' as the v4 implementation. I don't think it is necessary for this immediate PR, but maybe we can think about ways to make those clinically relevant categories pop out more (or maybe we just copy paste from v4)

For the drugs pages, I don't think our drugs will have associated categories. That is mostly a gene record associated data point (outside of immunotherapy, antineoplastic flags) so I think we can probably drop that from the drugs pages. For the 'Active' box, my original intent was to show some representation of market-active products that could be referenced easily for clinical use cases using the fields below:

Screen Shot 2023-01-19 at 8 43 10 AM

Originally that was intended to just be application numbers linked out to label information, but in its current form I am totally open to representing approval information there instead and revisiting labeling in a later issue/update.

mcannon068nw:
Implementing claims browsing seems like it might be a big lift and probably should be a future issue. We should brainstorm this going forward

jsstevenson (approved):

PR #259, opened 2023-01-20
refactor: use resolver method to avoid namespace collisions for :alias

jsstevenson:
Make these things go away:

```
DrugAlias's `field :alias` conflicts with a built-in method, use `resolver_method:` to pick a different resolver method for this field (for example, `resolver_method: :resolve_alias` and `def resolve_alias`). Or use `method_
conflict_warning: false` to suppress this warning.
```


acoffman (approved):
LGTM

PR #262, opened 2023-01-24
feat: add genes to browse category tables

### LINKED ISSUE #220: Browse Categories page uses dummy data
This page should use real data- may need to discuss with @jsstevenson or @acoffman where to retrieve this data

jsstevenson:
close #220 , was dummy data previously

TODO/maybe in new issues:
 * use paginated queries -- don't call for all genes at once
 * move over to MUI sorted tables
 * use correct counts for category listings -- currently provides claim counts (which may be > gene counts). I have a working query for this, I just don't know quite how to plug it in and I don't think it's a priority currently (was trying to get https://tanstack.com/query/v4/docs/react/guides/dependent-queries to work and didn't get it)
 * having trouble managing the ant panel/table component. When you open a different panel, rather than spawning with the loading animation, it'll retain whatever table was previously open. This is more noticeable for the really big categories. Also if you check/uncheck boxes on the side with a panel open, the panel won't update. Basically I have made a mess with the react state.


server/app/graphql/types/gene_claim_category_type.rb
```diff
@@ -4,19 +4,37 @@ class GeneClaimCategoryType < Types::BaseObject
...
     end
 
-    def gene_count
-      object.genes.count
+    field :genes, GeneCategoryResult.connection_type, null: false do
+      argument :source_names, [String], required: false
+      argument :category_name, String, required: true
+    end
+
+    def genes(source_names: [], category_name: '')
```
jsstevenson, line 24: @acoffman dumb question -- how do we reuse the `name` param passed to the type instead of providing it twice? nbd but it ends up making the query look a little silly:

```
    query geneClaimCategory($categoryName: String!, $sourceDbNames: [String!]!) {
      geneClaimCategory(name: $categoryName) {
        name
        genes(categoryName: $categoryName, sourceNames: $sourceDbNames) {
          edges {
            node {
              name
              conceptId
              longName
              sourceDbNames
            }
          }
        }
      }
    }
```

mcannon068nw:
found a weird behavior when testing locally that seems to be fairly replicatable. If I click on a large size category first, no data loads in (presumably until the large query is complete). If I click on a small category first, all data loads in fairly fast (even the larger query sets for some reason) no problem.

jsstevenson:
^^ I'm fairly confident that this is because we aren't paginating these queries. When the resulting data is large (IE every gene for a large category) that makes for a big query and a big response, hence long load times. We should paginate the query and incorporate paginated calls into the table component.

server/app/graphql/types/gene_claim_category_type.rb
acoffman: Are you asking how to avoid supplying `$categoryName` to both the `geneClaimCategory` and `genes` portions of the query?

That would be done in the back end resolver, first by having the `genes` resolver check if it has a parent or not and setting the default `scope` appropriately and then, depending on the structure of the query pulling the argument from its parent context. If that's unclear I'm happy to take a look at the code/make a PR

katiestahl (approved):

PR #274, opened 2023-02-01
Add drug name matching query

acoffman:
Essentially identical to the Gene matching query, run one query to bucket the search terms into match types and get the IDs of the matched drugs, then query the `drugs` endpoint with the list of IDs.

Example query:

```
{
  drugMatches(searchTerms: ["SUNITINIB", "ZALCITABINE", "LINCOCIN", "TRASTUZUMAB", "PHLORETIN", "NOTREAL"]) {
    directMatches {
      searchTerm
      matches {
        id
        name
      }
    }
    ambiguousMatches {
      searchTerm
      matches {
        id
        name
      }
    }
    noMatches {
      searchTerm
    }
  }
}
```

### DIFF
--- a/server/app/graphql/resolvers/drugs.rb
+++ b/server/app/graphql/resolvers/drugs.rb
@@ -9,7 +9,7 @@ class Resolvers::Drugs < GraphQL::Schema::Resolver
 
   scope { Drug.all }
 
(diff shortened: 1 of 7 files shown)


acoffman:
Definitely! I forgot we had those now

acoffman:
Test added!

jsstevenson (approved):
👍

PR #277, opened 2023-02-02
Fix drugs record table

### LINKED ISSUE #276: add PMIDs/Sources field to Drug Record page
related to #275, drug record page is missing PMID/Source information within interaction results table.

mcannon068nw:
closes #276 

Added sources/pmid retrieval to gql query and added back these fields to the individual record page.

### DIFF
--- a/client/src/hooks/queries/useGetInteractions.ts
+++ b/client/src/hooks/queries/useGetInteractions.ts
@@ -63,6 +63,13 @@ const getInteractionsByDrugsQuery = gql`
             type
             directionality
           }
+          publications {
+            pmid
+          }
+          sources {
+            id
+            fullName
+          }
         }
       }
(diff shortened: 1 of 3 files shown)


client/src/hooks/queries/useGetInteractions.ts
```diff
@@ -63,6 +63,13 @@ const getInteractionsByDrugsQuery = gql`
             type
             directionality
           }
+          publications {
+            pmid
+          }
+          sources {
+            id
```
jsstevenson, line 70: We shouldn't need to pull ID for anything on the frontend -- this can be removed

jsstevenson (changes requested):

jsstevenson (approved):

PR #285, opened 2023-02-09
Ambiguous terms UI

katiestahl:


jsstevenson:
👍 looks great. Two things I noticed (pls confirm that it's not just me):

- if you only search for an ambiguous match, it'll take you to the exact match screen with an unending loading animation, and then you click over to the ambiguous matches tab. IMO if there are ambiguous matches and no direct matches, you should go straight to the ambiguous tab (and the direct matches tab should be unclickable or otherwise able to indicate that no direct matches were found).
- the Antd "select # of results per page" dropdown isn't doing anything for me? IMO better to just make a switch over to MUI tables in the next few weeks, but if we don't get to it, we probably will need to fix this or take the option away

mcannon068nw:
> @jsstevenson good finds!
> 
> Regarding issue 1: I think we should create a new issue to handle that since it's sort of preexisting. I think it should be a relatively easy fix, though
> 
> And issue 2: you are correct- this page has never worked for any pages in DGIdb. I would like to convert the tables to MUI tables soon; I just haven't had the time to sit down and do that yet

for issue 2, what is the work/difficulty estimate on this? I am loving MUI so much more so would love this change but can probably make this a separate issue (I think)

jsstevenson (approved):

katiestahl:
Fixed James's suggested issue 1 in #286 
Table issues need to be addressed in #287

PR #296, opened 2023-02-22
feat: enable case-insensitive lookup

jsstevenson:


acoffman (approved):
LGTM

server/app/graphql/types/queries/drug_lookup_query.rb
```diff
@@ -2,20 +2,20 @@ module Types::Queries
...
 
         results = {
           direct_matches: [],
           ambiguous_matches: [],
           no_matches: []
         }
 
-        #find exact matches on drug name 
+        #find exact matches on drug name
         direct_symbol_matches = Drug.where(name: remaining_terms)
```
acoffman, line 19: Can drug names ever be mixed case in the DB?

jsstevenson: I had thought we enforced this constraint in the base importer method or the model, but it looks like we just upcase in each importer's call to create_drug_claim (this is sort of scary but apparently it's been working -- just checked `SELECT * FROM drugs WHERE upper(name) != name;` and there were no results)

PR #297, opened 2023-02-22
feat: add source metadata to backend

jsstevenson:
I will leave it up to the designers to determine how much of this should actually show up in the frontend, but this branch updates and provides more consistent formatting for all source citations and licensing, and breaks PMIDs/PMC IDs/DOIs into separate fields, so that they can be provided as separate links.


mcannon068nw (approved):
looks good to me

PR #298, opened 2023-02-24
Search persistence

### LINKED ISSUE #201: search terms get wiped out on refresh or returning to results from record page
a refresh should re-execute the search and not lose the search and end up on the landing page

katiestahl:
closes #201 
This PR introduces search persistence so a user does not need to re-submit their search after visiting a record page or refreshing on the results page

### DIFF
--- a/client/src/pages/Home/Home.tsx
+++ b/client/src/pages/Home/Home.tsx
@@ -22,14 +22,10 @@ export const Home: React.FC = () => {
 
   const handleSubmit = async () => {
     dispatch({type: ActionTypes.ContentPage})
-    if (state.interactionMode === 'categories') {
-      navigate('/categories');
-    } else {
-      navigate({
-        pathname: '/results',
-        search: `${createSearchParams({searchTerms: state.searchTerms.join(',')})}`,
-      });
-    }
+    navigate({
+      pathname: '/results',
+      search: `${createSearchParams({searchType: state.interactionMode, searchTerms: state.searchTerms.join(',')})}`,
+    });
   };
 
   const { refetch: refetchGenes } = useGetInteractionsByGenes(state.searchTerms);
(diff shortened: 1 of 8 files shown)


client/src/pages/Home/Home.tsx
```diff
@@ -2,33 +2,36 @@
...
 
   const handleSubmit = async () => {
     dispatch({type: ActionTypes.ContentPage})
     if (state.interactionMode === 'categories') {
       navigate('/categories');
     } else {
-      navigate('/results');
+      navigate({
+        pathname: '/results',
+        search: `?${createSearchParams({searchTerms: state.searchTerms.join(',')})}`,
```
acoffman, line 30: Gene symbols I believe will be alphanumeric, but concept ids may include `:` characters which would require URL encoding to work correctly here I believe, unless `createSearchParams` is handling that under the hood.

Also looking at the docs, I believe you may be able to omit the leading `?`

acoffman (commented):
Looks good overall to me, pending one small comment

katiestahl:
Just realized a potential issue- if users attempted to bookmark or share searches, we need some way of keeping track of what type of search they were doing. I'll look more into this to figure out the best way to keep track of this. I'm either thinking a flag or populating the search params based on the type of search (ie; "genes=" or "drugs=" instead of "searchTerms="

client/src/pages/Home/Home.tsx
katiestahl: ah yeah, was going to ask about this since I'm not sure all of the possibilities for gene/drug names. It appears that `createSearchParams` handles the encoding for us, which is nice!

mcannon068nw (approved):
built locally, remembers both search parameters and search type. Looks good! 👍

jsstevenson (approved):

PR #301, opened 2023-03-03
TSV Downloads of Table Contents

acoffman:
This introduces a new API endpoint and corresponding React component that allows for a user to download a data TSV based on the current parameters used in a query.

I wanted to get feedback on the approach before hooking it up to every table. It also will require some visual polish, I'm not sure how to constrain the size of the MUI loading spinner or error banner, so the button increases in size when either is present. (And the `LoadingButton` appears to be in beta/not shipped with the default library)

The `<TableDownloader/>` component takes two props: `tableName` which is the name of the table you want to download. This is registered in the  `table_resolvers` dictionary in the table downloads controller server side.

This determines which `Resolver` will get used to generate the TSV. (Basically which top level query is this table using).

Then, the resolver in question will need to implement two methods: `table_headers` and `to_row(object:)`. This instructs the resolver how to build the individual TSV rows. For `genes` here I've implemented the same columns we show on the frontend, but this gives us the flexibility to do other things too. We could include columns like ids, or urls to the individual records that we may not want to show visually in the table.

The second prop on the `<TableDownloader>` is `vars` This is simply the same data structure you pass to the `gql` query used to generate the table in the first place. This ensures the TSV is scoped in the same way as the table as the same params will be passed into the resolver. In this examples, it's the gene names used in the initial search query.

Finally, because the TSV can be arbitrarily large, the entire thing is streamed to the client by the server rather than being built up in memory and sent as one large chunk.

Its not a huge change in terms of lines of code, but its pretty dense, so I'm happy to walk through it start to finish on a call!

### DIFF
--- a/client/src/components/Gene/Categories/CategoryResults.tsx
+++ b/client/src/components/Gene/Categories/CategoryResults.tsx
@@ -9,6 +9,7 @@ import { GlobalClientContext } from 'stores/Global/GlobalClient';
 // styles
 import { Tabs } from 'antd';
 import './CategoryResults.scss';
+import TableDownloader from 'components/Shared/TableDownloader/TableDownloader';
 
 export const CategoryResults: React.FC = () => {
 
@@ -27,6 +28,7 @@ export const CategoryResults: React.FC = () => {
     <div className="category-results-container">
       <Tabs defaultActiveKey="1" onChange={onChange}>
         <TabPane tab="Unique Matches" key="1">
+        <TableDownloader tableName='gene_category_results' vars={{names: state.searchTerms}}/>
         <div className="gene-categories">
             {genes?.map((gene: any) => {
               return (
--- a/server/app/controllers/table_download_controller.rb
+++ b/server/app/controllers/table_download_controller.rb
@@ -1,19 +1,24 @@
 class TableDownloadController < ApplicationController
 
   @@table_resolvers = {
-    gene_interaction_results: Resolvers::Genes
+    gene_interaction_results: [Resolvers::Genes, GeneInteractionsTsv],
+    drug_interaction_results: [Resolvers::Drugs, DrugInteractionsTsv],
+    gene_category_results: [Resolvers::Genes, GeneCategoriesTsv]
   }
 
   def download_table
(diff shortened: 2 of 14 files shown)


mcannon068nw (approved):
from our call today, implementation seems very sound and amenable to any future updates we would need to do to resolvers, tsv mappings. Going to go ahead and approve pending removal of console log statements and anything else that @katiestahl or @jsstevenson recommend.

For future, we will need to consider how to best style the button still (alignment, etc.) as well as if we want to put a cap in the stream_table method

mcannon068nw:
per meeting today, @jsstevenson or @katiestahl would you be able to look through this PR as well just to make sure everything checks out?

server/app/controllers/table_download_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  private
+  def stream_table(resolver: , mapper:)
+    variables = prepare_variables(params[:variables])
+    variables.delete('sortBy')
+    variables.transform_keys! { |k| GraphQL::Schema::Member::BuildType.underscore(k) }
+
+    headers.delete("Content-Length")
+    headers["Cache-Control"] = "no-cache"
+    headers["Content-Type"] = "text/csv"
```
jsstevenson, line 28: we might be able to make this `text/tab-separated-values` but idk if being more specific has any benefits

client/src/components/Gene/Categories/CategoryResults.tsx
```diff
@@ -27,6 +28,7 @@ export const CategoryResults: React.FC = () => {
     <div className="category-results-container">
       <Tabs defaultActiveKey="1" onChange={onChange}>
         <TabPane tab="Unique Matches" key="1">
+        <TableDownloader tableName='gene_category_results' vars={{names: state.searchTerms}}/>
```
jsstevenson, line 31: There's an upstream issue we need to resolve to enable searching on more fields than just name (and to convert search terms to something unique, like concept ID). Once that's done, we'll need to remember to update this as well.

jsstevenson (approved):
looks great 👍

PR #302, opened 2023-03-05
Issue 199

rbasu101:


mcannon068nw:
merge conflicts fixed, ready for PR review

mcannon068nw:
Due to potential vulnerabilities, we should make this change even if i breaks the run example query buttons. If there is not a workaround for that, we can at the very least just display the sample queries and suggest copy/paste

client/package.json
```diff
@@ -26,12 +26,15 @@
     "create-react-app": "^5.0.0",
     "css-loader": "^6.7.1",
     "env-cmd": "^10.1.0",
+    "graphiql": "^2.4.0",
     "graphql": "^15.5.1",
+    "graphql-playground-react": "^1.7.28",
```
jsstevenson, line 31: if we're serving it all in the frontend now, we need to pull all of the stuff out of the backend

jsstevenson (changes requested):

client/src/pages/Playground/Playground.tsx
```diff
@@ -0,0 +1,192 @@
...
+      geneCategoriesWithSources {
+        name
+        sourceNames
+      }
+    }
+  }
+}
+`
+
+const fetcher = createGraphiQLFetcher({ url: 'http://localhost:3000/api/graphql'})
```
mcannon068nw, line 132: @acoffman @jsstevenson this works locally for me, but for deployment purposes, what should this point to? (I think there is likely an environment variable setting to use here but I do not know the best implementation)

acoffman: I would just import and use the `API_URL` const that's already available and pulled in from the env, that should work for you!

client/package.json
mcannon068nw: this change has been made

server/Gemfile
```diff
@@ -26,6 +26,7 @@ gem 'graphql', '~> 1.12.15'
 gem 'graphql-batch', '~> 0.4.3'
 gem 'graphiql-rails', '~> 1.7.0'
 gem 'search_object_graphql', '0.3.1'
+gem 'graphql-playground'
```
jsstevenson, line 29: we should be able to delete this stuff

jsstevenson (changes requested):

jsstevenson (approved):

PR #303, opened 2023-03-07
fix: make case insensitive search actually work

jsstevenson:


server/db/structure.sql
```diff
@@ -1880,3 +1880,4 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20230110183235'),
 ('20230214150515');
 
+
```
mcannon068nw, line 1883: this spacing is critical

mcannon068nw (approved):
can confirm, case insensitive search works for gene, drug, and categories with this PR

PR #319, opened 2023-03-21
Misc static info changes

### LINKED ISSUE #309: No versioning of sources on browse sources page
In v4.0, we have a concept of versioning for sources (see below) Screenshot 2023-03-14 at 10 21 04 AM
However, [here](https://staging.dgidb.org/browse/sources) we don't have the same concept of versioning.

mcannon068nw:
closes #309 

This PR makes a couple of small miscellaneous changes for the information pages:

- sources versioning: versioning has been added per source (tests updated to match)
- hyperlink sources: source headers now hyperlink out to their respective sites (tests updated to match)
- downloads: instructions for loading the v5 database have been updated, as well as a link to sql data dump

### DIFF
--- a/client/src/hooks/queries/useGetDruggableSources.ts
+++ b/client/src/hooks/queries/useGetDruggableSources.ts
@@ -14,6 +14,7 @@ query sources($sourceType: SourceTypeFilter) {
     nodes {
       sourceDbName
       sourceDbVersion
+      baseUrl
       categoriesInSource {
         name
         geneCount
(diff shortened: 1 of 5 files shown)


client/src/hooks/queries/useGetDruggableSources.ts
```diff
@@ -13,6 +13,8 @@ query sources($sourceType: SourceTypeFilter) {
     }
     nodes {
       sourceDbName
+      sourceDbVersion
```
jsstevenson, line 16: I think you're right that these queries are identical -- we should make a separate issue to clean it up

jsstevenson (changes requested):
small changes

jsstevenson (approved):

PR #342, opened 2023-04-13
fix: catch Pharos graphql client initialization errors

jsstevenson:
We've had some recurring issues with Pharos changing their API, or their servers going down entirely, which breaks the API client initialization and makes it hard to do... anything on the server (see eg below). This PR moves client initialization behind a `rescue` block which will hopefully mitigate some of the nuisance?

```
[ download_source_data] code/dgidb-v5/server % rails c
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
/Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/graphql-1.12.24/lib/graphql/schema/loader.rb:17:in `fetch': key not found: "data" (KeyError)
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/graphql-1.12.24/lib/graphql/schema/loader.rb:17:in `load'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/graphql-1.12.24/lib/graphql/schema.rb:769:in `from_introspection'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/graphql-client-0.18.0/lib/graphql/client.rb:52:in `load_schema'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/graphql-client-0.18.0/lib/graphql/client.rb:61:in `load_schema'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:30:in `<module:PharosApi>'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:22:in `<class:ApiClient>'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:6:in `<module:Pharos>'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:5:in `<module:ApiImporters>'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:5:in `<module:Importers>'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:5:in `<module:Genome>'
        from /Users/jss009/code/dgidb-v5/server/lib/genome/importers/api_importers/pharos/api_client.rb:5:in `<main>'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/zeitwerk-2.6.0/lib/zeitwerk/kernel.rb:35:in `require'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/activesupport-6.1.6/lib/active_support/dependencies.rb:332:in `block in require'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/activesupport-6.1.6/lib/active_support/dependencies.rb:299:in `load_dependency'
        from /Users/jss009/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/activesupport-6.1.6/lib/active_support/dependencies.rb:332:in `require'
        from /Users/jss009/code/dgidb-v5/server/config/initializers/genome.rb:1:in `block in <main>'
        from /Users/jss009/code/dgidb-v5/server/config/initializers/genome.rb:1:in `each'
        from /Users/jss009/code/dgidb-v5/server/config/initializers/genome.rb:1:in `<main>'
```

### DIFF
--- a/server/lib/genome/importers/api_importers/pharos/api_client.rb
+++ b/server/lib/genome/importers/api_importers/pharos/api_client.rb
@@ -27,25 +27,35 @@ module Genome; module Importers; module ApiImporters; module Pharos;
         end
       end
 
-      Schema = GraphQL::Client.load_schema(HTTP)
-      Client = GraphQL::Client.new(schema: Schema, execute: HTTP)
+      begin
+        Schema = GraphQL::Client.load_schema(HTTP)
+        Client = GraphQL::Client.new(schema: Schema, execute: HTTP)
+
+        Query = PharosApi::Client.parse <<-GRAPHQL
+          query($filter: IFilter, $skip: Int, $top: Int) {
+            targets(filter: $filter) {
+              targets(skip: $skip, top: $top) {
+                uniprot
+                name
+                sym
+                fam
+                preferredSymbol
+              }
+            }
+          }
+        GRAPHQL
+      rescue KeyError => e
+        puts "Error initializing Pharos GraphQL client: #{e.message}"
+        Schema = nil
+        Client = nil
+        Query = nil
+      end
     end
 
-    Query = PharosApi::Client.parse <<-GRAPHQL
-      query($filter: IFilter, $skip: Int, $top: Int) {
-        targets(filter: $filter) {
-          targets(skip: $skip, top: $top) {
-            uniprot
-            name
-            sym
-            fam
-            preferredSymbol
-          }
-        }
-      }
-    GRAPHQL
 
     def send_query(category, skip, top)
+      return [] unless PharosApi::Schema && PharosApi::Client && PharosApi::Query
+
       response = PharosApi::Client.query(Query, variables: {
         'filter': {
           'term': category


server/lib/genome/importers/api_importers/pharos/api_client.rb
```diff
@@ -27,25 +27,35 @@ def headers(_context)
...
+                preferredSymbol
+              }
+            }
           }
-        }
-      }
-    GRAPHQL
+        GRAPHQL
+      rescue KeyError => e
+        puts "Error initializing Pharos GraphQL client: #{e.message}"
```
jsstevenson, line 48: @acoffman is there a more helpful way to log this for the production environment?

acoffman: `Rails.logger.warn` or `Rails.logger.error` would be more canonical ways to log this, and then if you end up using logstash or cloudwatch or something it'll go to the right place instead of just stdout/stderr.

acoffman (approved):
This solution will work, but will require rebooting the app to re-initialize if it fails.

Is there a reason that the API client is bootstrapped in the module itself rather than in a method? 

I think I might put all the initialization code in a `def init` or `setup` method or similar that sets all the appropriate instance variables and just call that as the first step in enumerate genes or send query. That way it won't try to build the Pharos client at boot time, only the first time its actually used.

Either way, this will solve the problem for the immediate future.

jsstevenson:
> I think I might put all the initialization code in a def init or setup method or similar that sets all the appropriate instance variables and just call that as the first step in enumerate genes or send query. That way it won't try to build the Pharos client at boot time, only the first time its actually used.

maybe I'm missing something obvious here, but I tried a few things this morning and couldn't figure out how to make the schema/client/query variables re-initializeable without throwing [this error](https://github.com/github/graphql-client/blob/master/guides/dynamic-query-error.md). Any suggestions?

acoffman:
@jsstevenson Yeah, you're right I think. Something like this will work:

```ruby
require "graphql/client"
require "graphql/client/http"
require 'open-uri'

module Genome; module Importers; module ApiImporters; module Pharos;
  class ApiClient
    def enumerate_genes(category)
      skip = 0
      top = 50
      genes = send_query(category, skip, top)
      Enumerator.new do |y|
        until genes.empty?
          skip += genes.size
          genes.each { |gene| y << gene }
          genes = send_query(category, skip, top)
        end
      end
    end

    private

    module PharosApi
      def self.client
        init
        return Client
      end

      def self.ready?
        [
          defined?(HTTP),
          defined?(Schema),
          defined?(Client),
          defined?(Query),
        ].all?
      end

      def self.init
        if !defined?(HTTP)
          endpoint = 'https://pharos-api.ncats.io/graphql'
          const_set('HTTP', GraphQL::Client::HTTP.new(endpoint) do
            def headers(_context)
              { 'User-Agent': 'DGIdb.org Pharos importer' }
            end
          end)
        end

        begin
          if !defined?(Schema)
            const_set("Schema", GraphQL::Client.load_schema(HTTP))
          end

          if !defined?(Client)
            const_set('Client', GraphQL::Client.new(schema: Schema, execute: HTTP))
          end

          if !defined?(Query)
            const_set('Query', PharosApi::Client.parse(<<-GRAPHQL
              query($filter: IFilter, $skip: Int, $top: Int) {
                targets(filter: $filter) {
                  targets(skip: $skip, top: $top) {
                    uniprot
                    name
                    sym
                    fam
                    preferredSymbol
                  }
                }
              }
            GRAPHQL
           ))
          end

        rescue KeyError => e
          Rails.logger.warn("Error initializing Pharos GraphQL client: #{e.message}")
        end
      end
    end


    def send_query(category, skip, top)
      PharosApi.init
      return [] unless PharosApi.ready?

      response = PharosApi.client.query(PharosApi::Query, variables: {
        'filter': {
          'term': category
        },
        'skip': skip,
        'top': top
      })
      response.data.targets.targets
    end
  end
end; end; end; end

```

But I'm not sure the added indirection and complexity is worth it. Maybe its best to keep it as it is?

PR #347, opened 2023-04-25
Typeahead queries

acoffman:
This adds support on the back end to do left anchored name filtering by gene, category, and drug name, and then adds three new queries to the frontend to take advantage of it.

Now it should just be a matter of picking which of the three we want to use in the logic [here](https://github.com/dgidb/dgidb-v5/blob/dev/client/src/components/Shared/SearchBar/SearchBar.tsx#L72) and then hooking that up to the `Autocomplete` component. 

I'm happy to take this the rest of the way, but didn't want to stomp on any changes @katiestahl was already making in her search bar fixes branch.


jsstevenson:
👍 queries look good, could you add some basic test coverage?

katiestahl (approved):

acoffman:
Tests added as requested. Will merge after they pass on GitHub

PR #349, opened 2023-05-15
Search bar fixes

### LINKED ISSUE #34: Search page autocomplete
Implement autocomplete for genes + drugs

### LINKED ISSUE #221: transition from antd to mui?
The Wagner lab developers have more experience and familiarity with Material-UI over ant-design. Additionally, MUI provides for better customizability of components over ant-design. 

After some initial work and research, I estimate that transitioning to MUI components in the place of ant design ones will take around 2 weeks worth of work. I estimate that the _reduction in complexity_ for using MUI over ant-design for the remaining work required to produce a workable version of dgidb will lessen developer time spent on tickets as follows: (please note, this is simply an estimation)

3 point tickets - 1-2 days of time saved
5 point tickets - 3 days of time saved
8 point tickets - 5 days of time saved
13 points - 7 days of time saved

I came up with these estimations by analyzing some of the current pointed tickets that are in the backlog or in the sprint backlog, coming up with quick pseudo-designs, and determining the feasibility of implementing those designs with existing ant-d components vs implementing with existing MUI components. 

There still needs to be work done to determine an estimated point value of remaining work on dgidb v5; however, this will require a meeting of developers and stakeholders to determine exactly what requirements must be implemented to get to the target application for the manuscript release. 

Pros:

- reduced development time for tasks that the Wagner lab will be completing
- easier customizability for features that still need to be implemented
- cuts down on bundle size: antd's bundle size is huge compared to MUI, so doing a transfer to MUI and removing antd should show a performance improvement on slow networks on initial load?

Cons: 

- since civic uses antd, there may be some visual differences between using civic db or dgidb ( we need to determine if this is a deal-breaker)
- ~2 weeks of dev time spent toward the transition (but a pro: we can implement new features in MUI and gradually transition the old pages from antd to MUI, so we won't directly see this negatively impact development time anywhere)

Considerations:

- We could move forward with using both MUI and ant-design. There's not a big detriment here other than a larger bundle size. I would need to do some testing to determine how big the impact may be here, if any. Looks like MUI's bundle size is 490kB minified. This isn't negligible, but it's certainly far from antd's 1.2MB. 
- As long as we completely transition away from antd before launch, only developers or people accessing the dev instance would see any of this performance impact that could be caused by the large bundle size

From here on out, though, if I get the grace/go-ahead from @acoffman I would like to add material-ui to the dgidb v5 project and use Material-UI components over ant-design components for all features from here on out. Because of the nature of the app and developing it, we really need the customizability of components that is offered in MUI that ant-design is simply lacking.

### LINKED ISSUE #334: Typing in a search term and then clicking on one of the dropdown results doesn't populate the search term list
Using the keyboard to select a term works but not mouse select.

https://user-images.githubusercontent.com/6617192/229534354-6f6c74d2-8a67-43af-b1f1-a52d16c7c603.mov

katiestahl:
This PR encapsulates fixing numerous search bar bugs and converting components to MUI and removing antd from our stack.

Closes #34 
Closes #221 
Closes #334


jsstevenson (approved):

client/src/components/Shared/SearchBar/SearchBar.tsx
```diff
@@ -1,130 +1,134 @@
...
+            {...params}
+            variant="standard"
+            label="Search Terms"
+          />
+          </Box>
+        )}
+        value={selectedOptions}
+        onInputChange={handleInputChange}
+        onChange={handleAutocompleteChange}
+      />
```
jsstevenson, line 116:
```suggestion
        onInputChange={handleInputChange}
        onChange={handleAutocompleteChange}
        isOptionEqualToValue={(option, value) => option.name === value.name}
      />
```

@katiestahl this silences the warning

acoffman (approved):
This looks good overall to me!

I don't know if at some point we'd want to add a debounce to the typeahead so it doesn't fire on literally every keystroke, but that's something we can address later if needed.

PR #351, opened 2023-05-15
fix: don't use non-unique keys for URL params

jsstevenson:


mcannon068nw:
The way concept IDs are created, I think this will always be unique? Would we even need to set a constraint?

server/app/graphql/types/query_type.rb
```diff
@@ -52,11 +52,11 @@ def gene_claim(id:)
 
     field :gene, Types::GeneType, null: true do
       description "A gene"
-      argument :name, String, required: true
+      argument :conceptId, String, required: true
```
acoffman, line 55: Not a huge deal, but in ruby variables are generally `snake_case` (like Python), and in JS/Typescript, they're generally `camelCase`. To deal with that, if you make your variables `snake_case` here, the graphl runtime will actually automatically expose them as `camelCase` at the API/JS level: https://graphql-ruby.org/fields/arguments.html#automatic-camelization

So you can name this `concept_id` here to conform to ruby formatting standards, and still get `conceptId` on the client side "for free."

acoffman (approved):
One small nitpick but looks good to me!

jsstevenson:
@mcannon068nw I think it's unlikely that we'd ever accidentally submit duplicate concept IDs, but table-side SQL constraints are nice because we humans are error-prone and do unlikely things all the time

PR #354, opened 2023-05-23
feat: search on aliases

### LINKED ISSUE #304: Search by alias and concept ID
Not currently working. I think we want to use the ambiguous matches query to get concept IDs for a given set of search terms, and then populate tables and build pages based off of concept IDs, not names (since names aren't unique).

### LINKED ISSUE #353: use concept ID, not name/search term, in table download requests
not handling this in #351 because I think it will be much easier to do once the queries are cleaned up in the `search-alias` branch. The resolvers should get a new option a la

```ruby
option(:concept_ids, type: [String]) { |scope, value| scope.where("concept_id IN (?)", value) }
```

### LINKED ISSUE #357: Search categories autocomplete should suggest gene names

jsstevenson:
close #304
close #353 
close #357 

Enable autocomplete, search, etc on aliases and concept IDs. Combines a few queries and trims down the number of requested fields based on how we ultimately use them. Removes some unused frontend components and tried to whittle down some of the TS compiler warnings in the files I was working in. I also ran Prettier on some things that I probably shouldn't have so now there are a bunch of extra changes, sorry about that.

an open question (discuss?): some aliases look a little garbage-y, e.g.:

Screenshot 2023-05-24 at 10 23 34 AM

(these aren't necessarily that bad, but like, "BRAF V600E" shouldn't really be a synonym for "BRAF")

We currently drop redundant suggestions (i.e. so that the autocomplete dropdown isn't listing the same thing several times) but we could maybe do more things? 

* Don't provide multiple suggestions for the same concept ID
* Pull sourcing up to drug aliases and only provide suggestions from accepted sources
* Similar to the above, add a "suggestable" column to aliases and have some rules for setting that during grouping
* Pull nomenclature up from claim_aliases to aliases (and eg only suggest from "Trade name" "Generic name" "Alias" for drugs)


jsstevenson:
Reducing redundant suggestions by concept ID probably makes the more complex solutions unnecessary and would be pretty quick to implement. I think in a perfect world we could do something smarter but that'd be the most time-efficient improvement by far.

client/src/components/Gene/GeneSearchResults/GeneSearchResults.scss
```diff
@@ -0,0 +1,14 @@
+.ant-tabs-tab.ant-tabs-tab-active .ant-tabs-tab-btn {
```
acoffman, line 1: Same question as above

server/app/graphql/resolvers/drugs.rb
```diff
@@ -37,6 +37,10 @@ class Resolvers::Drugs < GraphQL::Schema::Resolver
     scope.where(concept_id: value)
   end
 
+  option(:concept_ids, type: String, description: 'Exact match filtering on concept IDs') do |scope, value|
+    scope.where('concept_id IN (?)', value)
```
acoffman, line 41: I'm a little confused at this one, `IN` is generally used for a list, and this argument is typed as a single string. Is it intended to be `[String]`. Additionally, doesn't matter a ton, but ActiveRecord is smart enough to handle an array with an `IN` for you, so if you change it, you can just pass in the array directly with `scope.where(concept_id: value)`.

server/app/graphql/types/query_type.rb
```diff
@@ -13,6 +13,98 @@ class QueryType < Types::BaseObject
...
+      description "A searchable drug name or alias that can be completed from the supplied term"
+      argument :term, String, required: true
+      argument :n, Int, required: false
+    end
+
+
+    def drug_suggestions(term:, n: 10)
+      matches = Drug.where("drugs.name ILIKE ?", "#{term}%").limit(n).map do |drug|
+        {
+          "drug_name": drug.name,
```
acoffman, line 26: Really small nitpick, the performance difference will be negligible in real life, but idiomatically you'd use symbols here for the dictionary keys, not strings, as each iteration will `new` up the strings again.

```ruby
{
   drug_name: drug.name,
   concept_id: drug.concept_id,
   suggestion: drug.name,
   suggestion_type: 'NAME'
}
```
and so on

acoffman (approved):
A couple of small nitpicks but overall it looks pretty good to me!

client/src/components/Drug/DrugSearchResults/DrugSearchResults.scss
```diff
@@ -0,0 +1,14 @@
+.ant-tabs-tab.ant-tabs-tab-active .ant-tabs-tab-btn {
```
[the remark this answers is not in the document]
jsstevenson: good catch, meant to circle back

server/app/graphql/types/query_type.rb
jsstevenson: Python brain at work, thanks

server/app/graphql/resolvers/drugs.rb
jsstevenson: Good catch, this is definitely a goof

PR #367, opened 2023-05-25
fix: handle messy CGI drug class names

### LINKED ISSUE #355: separate out drug class attributes from CGI
make this not happen:

Screenshot 2023-05-23 at 10 59 04 AM

jsstevenson:
close #355


mcannon068nw (approved):
🙌

PR #371, opened 2023-05-28
fix: make sure drug approval ratings are queried in search page

jsstevenson:
fixing what I broke


mcannon068nw (approved):
👍  looks good to me

PR #372, opened 2023-05-30
fix: clean CGI data

jsstevenson:


mcannon068nw (approved):
wow this looks a whole lot cleaner than what we had previously

PR #373, opened 2023-05-30
fix: add indication data to search results

### LINKED ISSUE #194: Missing indication field in results
In Interactions by Drug, certain records are missing an indication field. Might be a data issue?

jsstevenson:
close #194


mcannon068nw (changes requested):
this fixes for search by genes but does not seem to affect search by drugs

example case: CLOFARABINE for FLT3 shows "antineoplastic agent" for search by gene but not in search by drug

mcannon068nw (approved):
it is now working!

PR #377, opened 2023-05-31
fix: use buckets for approval rating graphic

### LINKED ISSUE #338: Drug approval graphic
Currently we have a visualization of regulatory approval on the interaction results page:

image

We should look into a color scheme that matches the ordinality of approval status / phase.

jsstevenson:
close #338 

Generally I think we probably have to use the same normalized values that show up in the results table -- otherwise it's a little confusing. It'd be great to put some thought into a combined approval rating strategy though (card in the drug results page/more ideas for how to display a summary in the interaction tables)

Screenshot 2023-05-31 at 1 09 56 PM


mcannon068nw (approved):
🙌

PR #379, opened 2023-05-31
fix: minor max_phase corrections

jsstevenson:
A while ago I refactored the ChEMBL importer and messed up a couple of small things. This won't affect a ton of claims but fixes a small infographic error.


mcannon068nw (approved):
👍

PR #380, opened 2023-06-01
feat: enable multi-filtering in summary and results listing

jsstevenson:
Fixes the filtering issue to adjust listed results in the interaction table. Also enables filtering on >1 gene/drug.


mcannon068nw (approved):
I love this PR. This works really well and I haven't found any bugs or anything running locally.

For a future thought, I wonder if the click name -> filter results/graph is intuitive enough, i.e. will a user know they can click it upon loading the page. If not, is there a way to communicate this?

PR #387, opened 2023-06-02
feat: provide hyperlinks to resources from xrefs

jsstevenson:
Screenshot 2023-06-02 at 12 23 24 PM
If an alias looks like a URI, try to generate a hyperlink out to the source page if it matches patterns for known resources.

It'd be great to restrict this behavior to a separate xrefs box on the drug/gene pages, but I think this is a good start


mcannon068nw (approved):
🥇 this looks great. We should be on the lookout for additional sources that we can link out (i.e. additional patterns to add)

PR #390, opened 2023-06-02
fix: address some out-of-date things in About pages

jsstevenson:
Needs at least one more pass -- the graphic/explanation has a lot of stuff about the query score, which we no longer provide (we should probably just redo the graphic in HTML)


client/src/pages/About/SubSections/Overview.tsx
```diff
@@ -27,20 +27,32 @@ export const Overview = () => {
...
+          target="_blank"
+        >
+          Therapy
+        </Link>{' '}
+        Normalizer services. DGIdb contains over 10,000 genes and 15,000 drugs
+        involved in over 50,000 drug-gene interactions or belonging to one of 43
+        potentially druggable gene categories. Users can enter a list of genes
+        to retrieve all known or potentially druggable genes in that list.
+        Results can be filtered by source, interaction type, or gene category.
+        DGIdb is built on Ruby on Rails and PostgreSQL with a flexible
```
mcannon068nw, line 54: Maybe we should mention something about GraphQL as well?

mcannon068nw (changes requested):

mcannon068nw:
looks good, request changes in lieu of a few content considerations

mcannon068nw (approved):

PR #392, opened 2023-06-05
fix: use symbol for CIViC gene claim name

jsstevenson:
Tiny fix, but with reason:

* in v4, we used gene symbols for gene claim names from CIViC. I think we switched for v5 because the CIViC API refers to the full name as `officialName`, and that sounds very official. But the Gene Normalizer can't currently search full names (we should probably fix this?), and anyway, we typically try to use the symbol for the claim name in all other sources.
* Incidentally, this would also fix our KRAS/NRAS problem.


acoffman (approved):
This looks like a good fix to me.

PR #393, opened 2023-06-06
feat: updated gene categories search results

### LINKED ISSUE #333: ambiguous/unmatched terms for gene categories search
I believe this should be implemented? or if not, the tab should be removed

jsstevenson:
close #333


mcannon068nw:
fwiw I think the updated layout in this draft looks great.

mcannon068nw (approved):
👍 switching to match cards is a good call

PR #400, opened 2023-06-12
Indication record bug

### LINKED ISSUE #399: gene/drug record pages do not load when there are no indications
On current dev branch, if you try navigating to a drug or gene record page with no available indications, the page stops rendering.

example: /drugs/chembl:CHEMBL346631

the bug seems to be happening due to a length > 0 line on the indications variable in InteractionTable/InteractionTable.tsx. I have made a bandaid fix on the branch 'indication-record-bug' using a try catch block, but there is definitely a better way to do this.

mcannon068nw:
Fixes #399


katiestahl (approved):
great job :)

PR #421, opened 2023-07-17
fix: correct source versioning

jsstevenson:
We were using the current date to set the version for a number of sources that are actually static imports (I think this was my bad). I went back and grabbed the date set in dgidb v4.


mcannon068nw (approved):
dates look good

PR #433, opened 2023-10-09
fix: update grouper API

jsstevenson:


server/lib/genome/groupers/drug_grouper.rb
```diff
@@ -40,28 +42,15 @@ def run(source_id = nil)
           if normalized_drug.is_a? String
             normalized_id = normalized_drug
           else
-            normalized_id = normalized_drug[@descriptor_name][@id_name]
-            create_new_drug(normalized_drug[@descriptor_name]) if Drug.find_by(concept_id: normalized_id).nil?
+            normalized_id = normalized_drug['normalized_id']
+            create_new_drug(normalized_drug['therapeutic_agent'], normalized_id) if Drug.find_by(concept_id: normalized_id).nil?
```
mcannon068nw, line 46: potentially dumb Q, but why the change to a hard coded 'therapeutic_agent' tag. Was the descriptor always therapeutic agent anyways?

mcannon068nw (approved):
lgtm.

Thank you for writing up the from scratch data load in the README

server/lib/genome/groupers/drug_grouper.rb
jsstevenson: This relates to some VA/VRS/GKS stuff introduced in the last few months. Previously, we were delivering VRSATILE therapeutic descriptors, but VRSATILE has been canned and therapeutics are a first class entity in the GA4GH knowledge model, e.g. described here: https://github.com/ga4gh/vrs-python/blob/1c9e4ab8d66c652a8d27430a77c4ad15834aa1ea/src/ga4gh/core/_internal/models.py#L229

The reason for instance variable vs hard coded string was there were some other methods that were able to be neutral with respect to drug versus gene if the `descriptor_name` instance var was available, but they kind of lost their utility over time and I think it's easier just to hard code things now

PR #455, opened 2023-12-01
feat: update API page to show users endpoint for graphql requests

### LINKED ISSUE #454: Add endpoint notice to API page
Currently the API page (https://www.dgidb.org/api) has no notice of how to programmatically access the endpoint outside of the playground functionality. Information should be added somewhere on this page to denote that queries can programmatically be sent to https://dgidb.org/api/graphql.

katiestahl:
close #454

- renamed Playground to API since I thought it might make a little more sense
- Added some general API information and headers to the page to display the url users should be making requests to

### DIFF
--- a/client/src/pages/API/API.scss
+++ b/client/src/pages/API/API.scss
@@ -1,3 +1,24 @@
+.api-page-header {
+  font-size: 36px;
+  margin-bottom: 15px;
+}
+
+.api-info-container {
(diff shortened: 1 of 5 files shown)


mcannon068nw (approved):
I think this will work for now and get the api endpoint info out, but we should probably see if we can come out with an additional layout for this page

PR #456, opened 2023-12-04
feat: add error message when data not found

### LINKED ISSUE #451: Untracked gene/drug page should display a 404-y message
> I noticed that if you navigate to a gene page that has an unsupported nomenclature
> https://beta.dgidb.org/genes/foo:391
> or even a supported one with an id that doesn't exist:
> https://beta.dgidb.org/genes/hgnc:3243434
> You get a standard results page, but everything is just empty. Do we think that should display some sort of 404 or "concept not found" or similar instead?

From @acoffman

katiestahl:
we should look into producing some sort of nice 404 error page with fancy dgidb branding and graphics down the road to make the app appeal more cohesive. 

but for now, this PR makes an error message appear instead of a weirdly populated page with a bunch of "no data found" messages

this also fixes the footer hanging out randomly in the middle of some pages where the content doesn't take up the whole page. It will always be pushed to the bottom

close #451


mcannon068nw (approved):
lgtm

agree that we should create a formal 404 page going forward

jsstevenson (approved):

client/public/images/dgidb-404-logo.png
jsstevenson, line 1: !!

PR #466, opened 2023-12-14
Setup analytics for v5

mcannon068nw:
The google analytics was automatically updated to GA4. Hooked in a basic implementation of this for DGIdb v5 that uses environmental variables to check for whether it should send data or not. This introduces a new package 'React-ga4'.

Currently tracks pageviews, sessions, and engagement. Would be nice to define and add in custom events for queries going forward.


mcannon068nw:
Actual modified files for this PR are: env.local, 
- .env.prod 
- .env.production 
- .env.staging
- App.tsx
- package.json

acoffman (approved):

PR #469, opened 2023-12-27
feat: make app more mobile friendly 

### LINKED ISSUE #461: make site more mobile friendly

katiestahl:
Opening this draft PR to help me keep track of my progress and what I still need to do. 

This PR will close #461 

Done so far:
- added hook to return boolean value if a screen size we consider "mobile" size is used
- made nav menu hamburger menu on small screen sizes + added a drawer and icons
- fixed majority of issues on the gene, drug, and interaction pages (stack each component vertically at 100% width)
- made components on home page not go off-screen 
- home page
- fixed bug where search results were getting repeated (not mobile-specific but needed a bandaid)
- point to ES6 because it's superior
- fixed about page (added floating button to jump between sections for mobile)
- browse sources - changed grid population to only display as many cards as possible given the space, otherwise it will wrap to next line. The noticeable change here will be that there will be more cards in a row for bigger screen sizes 
- make data tables fit nicely and scrollable on all screen sizes
- make results pages functional + look nice on all screen sizes
- reactive display for ambiguous/unmatched terms

Did not get to:
- browse categories - make drawer for selecting/deselecting categories?
- API page 
-

### DIFF
--- a/client/src/components/GeneCategories/AmbiguousMatchCard/AmbiguousMatchCard.tsx
+++ b/client/src/components/GeneCategories/AmbiguousMatchCard/AmbiguousMatchCard.tsx
@@ -67,7 +67,7 @@ export const AmbiguousMatchesCard: React.FC<Props> = ({ match }) => {
           </Box>
         ) : (
           <Typography fontStyle="italic">
-            STK1 is ambiguous. Please select context from drop down.
+            {match.searchTerm} is ambiguous. Please select context from drop down.
           </Typography>
         )}
       </AccordionDetails>
--- a/client/src/components/GeneCategories/AmbiguousMatchCard/AmbiguousMatchCard.tsx
+++ b/client/src/components/GeneCategories/AmbiguousMatchCard/AmbiguousMatchCard.tsx
@@ -67,7 +67,8 @@ export const AmbiguousMatchesCard: React.FC<Props> = ({ match }) => {
           </Box>
         ) : (
           <Typography fontStyle="italic">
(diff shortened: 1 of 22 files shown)


client/src/components/GeneCategories/AmbiguousMatchCard/AmbiguousMatchCard.tsx
```diff
@@ -67,7 +67,8 @@ export const AmbiguousMatchesCard: React.FC<Props> = ({ match }) => {
           </Box>
         ) : (
           <Typography fontStyle="italic">
-            STK1 is ambiguous. Please select context from drop down.
```
katiestahl, line 70: lol i can't believe this has been in there for as long as it has

mcannon068nw (approved):

jsstevenson:
Looking great overall, this is awesome. A few thoughts that could be addressed here or broken out to separate issues

- I think the logo on the top bar should downsize a bit when the menu switches to the hamburger. That would reduce the height of the top bar substantially which gives us more screen real estate
- The "current contributors" section in the About page could probably collapse to one column once we're in mobile-width territory
- I think it might look cleaner to make the API/downloads/GitHub links go vertically-stacked when width is minimal. Otherwise they look slightly offcenter from the text above:

Screenshot 2024-01-10 at 4 48 25 PM

- It would be great if we could make this indentation flush with the header (also would be nice if the bottom of the download button was vertically aligned with the bottom of the number text:

Screenshot 2024-01-10 at 4 49 53 PM

jsstevenson (approved):

PR #482, opened 2024-02-16
Add support for GA4GH service info query

### LINKED ISSUE #481: add dgidb version check to api
From a user suggestion, it would be useful to be able to check the overall DGIdb version via the API. This would allow users to quickly see if they need to update local data or are able to skip this resource due to no new data.

acoffman:
This adds an initializer in `config/initializers` that will run at boot. It expects to find a `data_version.yml` at the root dir of the application. That file should contain a key `version` containing the current data version.

This will be loaded into a constant.

When generating a new data release, you will need to remember to update this value (so, if there are docs for the process somewhere, may be good to add this step to them). Its manual for now, but as we automate the data generation process, we should automate this too.

Additionally, I have added a top level `serviceInfo` query to the API containing all the fields needed for the [GA4GH Service discovery spec](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ga4gh-discovery/ga4gh-service-info/develop/service-info.yaml). Though the query obviously looks different given it is GraphQL rather than REST, the response should be compliant with their schema.

`updated_at` and `version` information for this is pulled live from GitHub. (`dataVersion` is still handled as described above)

The query
```gql
{
  serviceInfo {
    id
    name
    type {
      group
      artifact
      version
    }
    description
    organization {
      name
      url
    }
    contactUrl
    documentationUrl
    createdAt
    updatedAt
    environment
    version
    dataVersion
  }
}

```
Will yield
```json
{
  "data": {
    "serviceInfo": {
      "id": "org.dgidb.graphql",
      "name": "DGIdb",
      "type": {
        "group": "org.dgidb",
        "artifact": "DGIdb GraphQL",
        "version": "v.5.0.4"
      },
      "description": "An open-source search engine for drug-gene interactions and the druggable genome.",
      "organization": {
        "name": "Wagner and Griffith laboratories",
        "url": "https://dgidb.org/about#contact"
      },
      "contactUrl": "mailto:help@dgidb.org",
      "documentationUrl": "https://dgidb.org/api/graphiql",
      "createdAt": "2023-10-20T08:51:00-05:00",
      "updatedAt": "2024-01-30T18:04:04+00:00",
      "environment": "development",
      "version": "v.5.0.4",
      "dataVersion": "Dec-2023"
    }
  }
}
```

closes #481

### DIFF
--- a/client/src/pages/About/SubSections/Contact.tsx
+++ b/client/src/pages/About/SubSections/Contact.tsx
@@ -6,7 +6,7 @@ export const Contact = () => {
     <div className="contact-section-container doc-section">
       <div>
         <p>
-          DGIdb was developed at The McDonnell Genome Institute, Washington
+          DGIdb was initially developed at The McDonnell Genome Institute, Washington
           University School of Medicine. If you have a source of information
           related to the druggable genome you would like us to incorporate,
           please contact us at{' '}
@@ -25,11 +25,11 @@ export const Contact = () => {
       <div className="left-section">
         <h4>
           <Link
-            href="http://genome.wustl.edu/"
+            href="https://griffithlab.org"
             target="_blank"
             rel="noreferrer"
           >
-            The McDonnell Genome Institute
+            The Griffith Laboratory
           </Link>
         </h4>
         <p>Washington University</p>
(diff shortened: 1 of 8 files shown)


ahwagner (approved):
LGTM 👍

PR #493, opened 2024-04-23
fix: use latest CIViC GraphQL API and capture all genes

jsstevenson:
* Fix API breakage from latest CIViC GraphQL schema update
* Rewrite query strategy to first capture all drugs/genes, then all available interactions between them, rather than only importing drugs/genes that are involved in predictive statements. This enables us to eg link out to more CIViC entities than we would otherwise.

### DIFF
--- a/server/lib/genome/importers/api_importers/civic/api_client.rb
+++ b/server/lib/genome/importers/api_importers/civic/api_client.rb
@@ -4,12 +4,32 @@ require "graphql/client/http"
 module Genome; module Importers; module ApiImporters; module Civic
   class ApiClient
 
-    def enumerate_variants
-      variant_edges = send_query
+    def enumerate_drugs
+      drug_edges = send_query(DrugsQuery).therapies.edges
       Enumerator.new do |y|
-        until variant_edges.empty?
-          variant_edges.each { |edge| y << edge }
-          variant_edges = send_query(variant_edges[-1].cursor)
+        until drug_edges.empty?
+          drug_edges.each { |edge| y << edge.node }
+          drug_edges = send_query(DrugsQuery, drug_edges[-1].cursor).therapies.edges
+        end
+      end
+    end
+
+    def enumerate_genes
+      gene_edges = send_query(GenesQuery).genes.edges
+      Enumerator.new do |y|
+        until gene_edges.empty?
+          gene_edges.each { |edge| y << edge.node }
+          gene_edges = send_query(GenesQuery, gene_edges[-1].cursor).genes.edges
+        end
+      end
+    end
+
+    def enumerate_evidence_items
+      evidence_items = send_query(InteractionsQuery).evidence_items.edges
+      Enumerator.new do |y|
+        until evidence_items.empty?
+          evidence_items.each { |edge| y << edge.node }
+          evidence_items = send_query(InteractionsQuery, evidence_items[-1].cursor).evidence_items.edges
         end
       end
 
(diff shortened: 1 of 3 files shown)


server/lib/genome/importers/api_importers/civic/api_client.rb
```diff
@@ -4,12 +4,32 @@
...
 
-    def enumerate_variants
-      variant_edges = send_query
+    def enumerate_drugs
+      drug_edges = send_query(DrugsQuery).therapies.edges
       Enumerator.new do |y|
-        until variant_edges.empty?
-          variant_edges.each { |edge| y << edge }
-          variant_edges = send_query(variant_edges[-1].cursor)
+        until drug_edges.empty?
```
acoffman, line 10: This works just fine as written, so no need to change it unless you want to. But just FYI, you can also request 

```
    pageInfo {
      hasNextPage
      endCursor
    }
```

At the top level of any of these queries. This avoids bringing the cursor back for each item, and just gives you the cursor of the last one in the page, and then provides a boolean `hasNextPage`. That way you can set up a loop sort of like (pseudocode) `while hasNextPage == true; send_query(after: endCursor)`.  Which will save you having to send a final query which comes back empty.

The overhead of the extra cursors and the one additional request is minimal so I don't think its worth rewriting, just something to know for next time.

server/lib/genome/importers/api_importers/civic/importer.rb
```diff
@@ -9,14 +9,87 @@ class Importer < Genome::Importers::Base
...
           end
 
           private
 
+          # this should be rare -- maybe impossible under the latest schema changes?
+          def importable_drug?(drug)
+            drug.name.upcase != 'N/A' && !drug.name.include?(';')
+          end
+
+          def importable_evidence_item?(ei)
```
acoffman, line 28: I think maybe the logic here got swapped around but the name didn't. For instance, wouldn't we want to import evidence with a rating `>= 2`, not `<=`? 

Unless I'm reading it wrong, the [old method](https://github.com/dgidb/dgidb-v5/pull/493/files#diff-7f1159fd04b53c023a0e172db1c5628bdfe738522e5b246e74a831c103aea9a5L41) maybe looks closer to correct?

This look more like its checking if we should skip the evidence, rather than import it.

server/lib/genome/importers/api_importers/civic/importer.rb
```diff
@@ -9,14 +9,87 @@ class Importer < Genome::Importers::Base
...
+          end
+
+          # current policy is to skip items with poor rating/level, anything that doesn't support the claim,
+          # or any item with >1 genes (e.g. a fusion)
+          def create_interaction_claims
+            @api_client.enumerate_evidence_items.each do |ei|
+              next if !importable_evidence_item?(ei)
+
+              # retain molecular profiles consisting of multiple variations on the same gene
+              gene_names = ei.molecular_profile.variants.map { |v| v.feature.feature_instance.name.upcase }.uniq
```
acoffman, line 73: A `feature` is 1:1 with a `featureInstance`. Essentially a `feature` exists in the schema where `gene` used to be, and then `featureInstance` is where any fields specific to the given feature type (ie: entrez id for a Gene, ncit id for a Factor, etc).

This a longwinded way of saying that any variants that share a feature_id share the same gene (assuming the feature is in fact a gene), so if you wanted to just make sure all the variants had the same feature id, that might be more straightfoward than name matching.

If you want to exclude non-gene variants, you could pull back the `__typename` of the feature instance, make sure its `Gene`. If that's true and all the ids match, you shouldn't need to do any further comparisons.

acoffman (changes requested):
Looks good to me overall! A couple of small comments that you can take or leave. The EID import logic is the only thing I think actually needs "fixed" unless I'm reading it wrong.

server/lib/genome/importers/api_importers/civic/importer.rb
jsstevenson: Yup, I flipped around the logic but got tripped up. I think it's simpler the old way.

acoffman (approved):

server/lib/genome/importers/api_importers/civic/api_client.rb
jsstevenson: @acoffman am I misunderstanding how this works? this query gives me `"hasNextPage": false` in the playground even though I assume there are > 27 drugs

```
query {
  therapies(first: 26, after: "") {
    pageInfo {
      endCursor
      hasNextPage
    }
    edges {
      node {
        id
        name
        ncitId
      }
    }
  }
}
```

acoffman: @jsstevenson You do understand it correctly, turns out there is actually a small issue with the SQL powering this specific query. I have a fix ready and it should go out today or tomorrow. Sorry for the inconvenience!

jsstevenson: (what a relief, figured it was either that or i was going insane)

Issue #257, opened 2023-01-19
Fix drug approval test case


jsstevenson:
As noted [here](https://github.com/dgidb/dgidb-v5/pull/248), we have one weird case in the backend tests where a drug refuses to allow `approved` to be set to the non-default `true`.

PR #494, opened 2024-04-24
feat: support enhanced filtering on interactions query

### LINKED ISSUE #490: Add arguments to `interaction` graphql query
Something like

* gene/drug name
* gene/drug term
* gene/drug concept ID
* filters (MOA? etc)

jsstevenson:
close #490 

Create an `interactions` query that lets the user constrain the universe of drugs/genes to search for interactions within. Also provides further drug filtering on the usual booleans.

* In trying to write tests for the latter, I hit an old issue that it seems impossible to make a `drug` from FactoryBot that has any of the antineoplastic/immunotherapy/approved flags set to `true`. I was thinking it had to do with the way those flags are initially set when the model instance is created, but this only confused me further. It's written up in #257 , still not sure what the deal is.


mcannon068nw (approved):
I think this is good as is. I can't think of any examples now, but we should keep an eye out for creative/interesting ways to query interaction data going forward that we haven't thought of.

PR #499, opened 2024-05-21
fix: exclude factors from unique gene name check in MPs

jsstevenson:


server/lib/genome/importers/api_importers/civic/importer.rb
```diff
@@ -71,7 +71,14 @@ def create_interaction_claims
 
               # retain molecular profiles consisting of multiple variations on the same gene,
               # but skip multi-gene profiles (eg fusions)
-              gene_names = ei.molecular_profile.variants.map { |v| v.feature.feature_instance.name.upcase }.uniq
+              gene_names = ei.molecular_profile.variants.map do |variant|
+                feature_instance = variant.feature.feature_instance
+                if feature_instance.respond_to?(:name)
```
acoffman, line 76: This looks like it should work, but it seems potentially a bit fragile given that factors (and all other feature types) will also have names, and this conditional working is merely a side effect of how the query happens to be written. Looking at the query [here](https://github.com/dgidb/dgidb-v5/blob/3f324e96d79ae1a8ca23f0ff2999fe0245d5c990/server/lib/genome/importers/api_importers/civic/api_client.rb#L111) I wonder if it should be changed to something more like 

```
feature {
    name
    featureInstance{
      __typename
    }
}
```

and then you could change the condition to just check the type and either filter out Factors explicitly, or limit your map to only variants where typename is Gene. Not critical either way, just a thought.

mcannon068nw (approved):
👍

server/lib/genome/importers/api_importers/civic/importer.rb
jsstevenson: @acoffman is there a way to form the graphql query to just skip features that aren't of type Gene? I'd prefer that over including some kind of discriminant in the response (that was what I thought I was originally doing, to be honest, but clearly it didn't work...)

acoffman: In this particular context... not really as you're going through the Molecular Profile.

That would lead to semi weird results if say you had a molecular profile consisting of two variants: `GENE VARIANT1 AND NOT FACTOR VARIANT2`, if you omit the Factor variant from that profile, you've fundamentally changed the meaning of the MP (and potentially evidence attached to it) if that makes sense?

In other contexts like the browse variants or browse features queries you definitely can just filter on variant/feature type.

PR #500, opened 2024-05-22
feat: make 404 page global error page and have unrecognized paths rou…

### LINKED ISSUE #396: Global error page

### LINKED ISSUE #460: Unknown routes should yield 404 errors, not redirect to home
Maybe this is intentional, but it strikes me as suboptimal in terms of giving meaningful feedback

katiestahl:
…te to it

close #396 
close #460

### DIFF
--- a/client/src/components/Drug/DrugRecord/DrugRecord.tsx
+++ b/client/src/components/Drug/DrugRecord/DrugRecord.tsx
@@ -277,6 +277,6 @@ export const DrugRecord: React.FC = () => {
       </Box>
     </Box>
   ) : (
-      <NotFoundError errorMessage='We could not find any results for this drug.' />
+    <NotFoundError errorMessage="We could not find any results for this drug." />
   );
 };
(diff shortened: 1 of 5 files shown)


jsstevenson (approved):
Also approving this in principle but a few things stuck out to me that might be worth thinking about

1) the <a> boxes for the links extend the entire width of the page for some reason? so if you click in the whitespace way to the right it still clicks a link. I think I'd expect the anchor to be on the text, not on the div surrounding the text
Screenshot 2024-05-25 at 9 39 09 PM

2) how would it look if the back/home links were also in the red box, underneath the icon/message?

3) I don't think a 404 HTTP response status gets sent -- that should probably be happening, assuming it's possible

side note: look at how fun the UniProt 404 icon is: https://www.uniprot.org/proteomdes?query=*

mcannon068nw:
This looks good as it. I do think its a little odd that the entire length of the line is clickable as a link but I don't think that it make or breaks this fix.

mcannon068nw (approved):

PR #501, opened 2024-05-22
feat: adding breadcrumbs for users to easily navigate back to search …

### LINKED ISSUE #339: Add titles to pages and/or breadcrumbs bar
It would be good to have the name of a page (or a breadcrumbs bar) on pages to better help a user understand where they are + assist with navigation

katie-perry:
…results on record pages

I had to adjust some styling for this. The 404 Not Found page may look a little silly, but I will fix it up and add a container for it once #500 is merged

close #339

### DIFF
--- a/client/src/components/Browse/Categories/BrowseCategories.scss
+++ b/client/src/components/Browse/Categories/BrowseCategories.scss
@@ -1,44 +1,10 @@
 .browse-categories-container {
+  background-color: var(--background);
+  padding: 10px;
   display: flex;
   flex: stretch;
 
   .category-list {
     flex: 1;
-  }
-
-  .ant-collapse-content {
-    .ant-collapse-content-box {
-      padding: 0px !important;
-    }
-  }
-
-  .ant-table-wrapper {
-    background-color: var(--background-light) !important;
-  }
-
-  .ant-table-pagination .ant-pagination {
-    margin: 8px 0 !important;
-  }
-
-  .ant-table-container table > thead > tr > th {
-    background-color: var(--background-light);
-  }
-}
-
-.ant-checkbox-group {
-  display: flex;
-  flex-direction: column;
-}
-
-.ant-checkbox-group-item,
-.ant-checkbox-wrapper {
-  color: var(--text-content);
-}
-
-.ant-collapse {
-  background-color: var(--background-content);
-
-  .ant-collapse-header {
-    color: var(--text-content) !important;
   }
 }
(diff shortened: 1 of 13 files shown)


jsstevenson (commented):
This looks really nice!

A couple cases for which I'm not sure is the right way to handle things:

1) if I keep clicking different drugs and genes, it'll send me down a chain of record pages, but "Search Results" still points back to the original search. I think that's correct, and the alternative would be like a growing chain of random drugs and genes (complicated to retain, could get too big) but it's a little unexpected also

2) If you click on a gene in Browse Categories, the breadcrumb is still "Search Results", and it points to the last search you did in that session. It would be really nice if we could set ourselves up to ensure the breadcrumb goes back to Browse Categories, and if the link opened the tab for the category that you clicked from (would probably entail some non-minor lifting on that page though).

3) If you are looking at a gene/drug page, and then put in the URL for a different gene/drug page, the breadcrumb will point back to your previous search results (even though you technically didn't get there via searching). Opening a new tab fixes this, but I'd say it's unexpected behavior.

katiestahl:
> This looks really nice!
> 
> A couple cases for which I'm not sure is the right way to handle things:
> 
> 1. if I keep clicking different drugs and genes, it'll send me down a chain of record pages, but "Search Results" still points back to the original search. I think that's correct, and the alternative would be like a growing chain of random drugs and genes (complicated to retain, could get too big) but it's a little unexpected also
> 2. If you click on a gene in Browse Categories, the breadcrumb is still "Search Results", and it points to the last search you did in that session. It would be really nice if we could set ourselves up to ensure the breadcrumb goes back to Browse Categories, and if the link opened the tab for the category that you clicked from (would probably entail some non-minor lifting on that page though).
> 3. If you are looking at a gene/drug page, and then put in the URL for a different gene/drug page, the breadcrumb will point back to your previous search results (even though you technically didn't get there via searching). Opening a new tab fixes this, but I'd say it's unexpected behavior.

thank you so much for testing this thoroughly and leaving this amazing feedback!! 

1. Yeah, I was trying to think about how to correctly represent the breadcrumbs for this page. I agree it is a bit unexpected. I'll do a bit more research to see what the standard is here and make an update. 
2. oh dang, I definitely missed this case. I'm wondering if it might make sense to just keep a state of previous pages visited and popping them off the array as the user goes back... 
3. Ah, I agree. I'm going to do some more research to figure out the best way to handle these scenarios

mcannon068nw (changes requested):
I agree that this looks really nice! We are missing one common use case though, and that is for interaction pages. 

Currently if you click on an interaction table row to navigate to an interaction info page, there is no breadcrumbs on that page.

Issue #515, opened 2024-08-01
Report Interaction Score components


paulflang:
IIUC `Interaction Score = evidence score * relative drug specificity * relative gene specificity`. Unfortunately, I can only find the Interaction Score, but not it's components in the Downloads section. Would it be possible to report the Interaction Score along with its three components? This would help us for filtering interactions.

mcannon068nw:
Thank you! We agree these could be useful and informative data for how the interaction scores are calculated. We are looking into potentially including the calculated components alongside the interaction score in future data downloads.

jsstevenson:
closed in #523

paulflang:
Fantastic. Thank you!

mcannon068nw:
hello @paulflang! I just wanted to reach out and let you know that these reported components are still coming. We ran into an unexpected data import issue while preparing the latest update. When this is resolved, new TSVs with these components reported will be released alongside the data update.

paulflang:
Sounds good. Thank you very much for keeping me in the loop.

PR #518, opened 2024-08-27
feat: store score components in DB

jsstevenson:
support #515

### DIFF
--- a/server/app/models/interaction.rb
+++ b/server/app/models/interaction.rb
@@ -68,7 +68,7 @@ class Interaction < ActiveRecord::Base
     interaction_score = evidence_score * drug_specificity * gene_specificity
 
     if update
-      self.interaction_score = interaction_score
+      self.score = interaction_score
       self.drug_specificity = drug_specificity
       self.gene_specificity = gene_specificity
       self.evidence_score = evidence_score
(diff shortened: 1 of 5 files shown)


jsstevenson:
add to graphql type

server/app/models/interaction.rb
```diff
@@ -51,15 +51,29 @@ def interaction_score(known_drug_partners_per_gene = nil, known_gene_partners_pe
...
+    if update
+      self.score = interaction_score
+      self.drug_specificity = drug_specificity
+      self.gene_specificity = gene_specificity
+      self.evidence_score = evidence_score
+      save!
+    end
+
+    interaction_score
   end
```
mcannon068nw, line 77: this looks right, at least as far as https://dgidb.org/about#interaction-scores is concerned. I would like to get @acoffman just to confirm that the ruby isn't doing anything unexpected here

mcannon068nw (approved):

server/app/models/interaction.rb
acoffman: This looks correct to me on first glance. Only "gotcha" I could think of is defaulting to integer division between two ints but it looks like that's covered by the `.to_f` calls.

PR #522, opened 2024-10-01
feat: jax-ckb -> ckb-core

jsstevenson:


mcannon068nw:
naming scheme seems ok. leaves room for and differentiates from CKB-Boost if we some how get access to that

mcannon068nw (approved):

PR #526, opened 2024-10-14
feat: log graphql requests

jsstevenson:
This PR does two things

* Logs query types and variables for things that you type into the search bar, basically.
* Logs the complete query and variables for requests made to the API endpoint NOT from the frontend client.

For example, the frontend logs look like:

```
# Logfile created on 2024-10-24 15:35:22 -0400 by logger.rb/v1.5.0
I, [2024-10-24T16:24:18.135312 #77300]  INFO -- : {:user_ip=>nil, :query_type=>"drugMatches", :query_variables=>{"searchTerms"=>["BRAF", "NTRK1", "NTRK2"]}}
I, [2024-10-24T16:24:25.670429 #77300]  INFO -- : {:user_ip=>nil, :query_type=>"drugMatches", :query_variables=>{"searchTerms"=>["IMATINIB"]}}
```

and the API logs look like

```
# Logfile created on 2024-10-24 15:35:22 -0400 by logger.rb/v1.5.0
I, [2024-10-24T16:20:56.239529 #76634]  INFO -- : {:user_ip=>nil, :query=>"query interaction($id: ID!) { interaction(id: $id) { gene { name conceptId } drug { name conceptId } interactionTypes { directionality type } interactionScore publications { id pmid citation } sources { fullName } interactionAttributes { name value } } }", :query_variables=>{"id"=>"a80d7cbb-d350-45c9-9800-6044db8dace0"}}
I, [2024-10-24T16:20:58.727886 #76634]  INFO -- : {:user_ip=>nil, :query=>"query gene($conceptId: String!) { gene(conceptId: $conceptId) { conceptId name geneAttributes { name value } geneAliases { alias } geneClaims { source { citation } } geneCategories { name } } }", :query_variables=>{"conceptId"=>"hgnc:3765"}}
```

Similarly, I was thinking it'd be nice to grab a reduced list of ambiguous results when those come up, but I haven't had a chance to look through the way to explore the query tree and I'm sort of supposed to get this thing shipped.


server/app/jobs/submit_analytics_event.rb
```diff
@@ -0,0 +1,21 @@
+require 'uri'
+require 'net/http'
+
+class SubmitAnalyticsEvent < ApplicationJob
+  # TODO not entirely sure what goes here
```
jsstevenson, line 5: ^

client/src/hooks/queries/useGetAmbiguousResults.ts
```diff
@@ -112,9 +112,13 @@ export function useGetMatchedResults(names: string[], type: ResultTypes) {
...
     async () => {
-      const res = await graphQLClient.request(requestQuery, {
-        searchTerms: names,
-      });
+      const res = await graphQLClient.request(
+        requestQuery,
+        {
+          searchTerms: names,
+        },
+        { 'dgidb-client-name': 'dgidb-frontend' }
```
acoffman, line 120: You might want to double check that it works as expected, but it looks like (based on this example [here](https://github.com/graffle-js/graffle/blob/graphql-request/examples/legacy/request-authentication-via-http-header.ts#L5)) you can set an HTTP Header when your initialize the graphql client.

That might be better than stapling it into each query individually.

jsstevenson: @acoffman yeah, I was a little unsure of whether we wanted to set different headers for different queries and to filter out noisier or less informative ones. We should probably set a base one at client initialization and then maybe use an additional one for chosen queries if we go that route.

server/app/lib/frontend_analytics_tracer.rb
```diff
@@ -0,0 +1,13 @@
+module FrontendAnalyticsTracer
+  FRONTEND_LOGGER = Logger.new(File.join(Rails.root, 'log', 'frontend-queries.log'))
+  def analyze_query(query:)
+    params = {
+      user_ip: query.context[:request_ip],
```
acoffman, line 5: You asked in the PR

>I have the client rigged to send different HTTP headers for each but I don't know if I can access that at the level of the tracer?

You can access anything that you stick in the query context in the controller like we're doing with IP here.
So if you want to pull out your header in the controller, put it in the context, and then pull it out in one of your tracers, that will work!

acoffman (changes requested):
A couple small things, but its looking good overall!

server/app/lib/frontend_analytics_tracer.rb
jsstevenson: Oh.... yeah, duh

server/app/lib/api_analytics_tracer.rb
```diff
@@ -0,0 +1,17 @@
+module ApiAnalyticsTracer
+  API_LOGGER = Logger.new(File.join(Rails.root, 'log', 'api-queries.log'))
+  def analyze_query(query: )
+    query_type = query.selected_operation.selections&.first&.name
+
+    return unless query_type == '__schema'
```
acoffman, line 6: I think this needs to be inverted right? We skip analytics if it is an introspection query.

acoffman (changes requested):
One minor question, looks good overall!

server/app/lib/api_analytics_tracer.rb
jsstevenson: Ugh, thanks, I always goof these

acoffman (approved):
LGTM!

PR #528, opened 2024-10-15
build: remove some unused rails boilerplate/imports

jsstevenson:
Not sure how kosher this is, but I went ahead and removed some boilerplate comments/imports that we didn't appear to be using. Let me know if anything looks problematic.


acoffman (approved):
Seems good to me!

PR #529, opened 2024-10-22
fix: remove redundant code

jsstevenson:
A little cleanup of unused/redundant stuff -- maybe a merge artifact?


mcannon068nw (approved):
surprised this made it into the Dev branch. App won't launch with the duplicated code to begin with. Will look out for this more diligently in the future

katiestahl (approved):
I'm surprised this didn't fail or get caught by the lint checks for before we merge. Might need to revisit that. I bet this was caused by a merge conflict though

PR #531, opened 2024-11-11
feat: log API client

jsstevenson:
The intention here is that we can tell if `dgipy` is being used to issue requests against the API

### DIFF
--- a/server/app/lib/api_analytics_tracer.rb
+++ b/server/app/lib/api_analytics_tracer.rb
@@ -9,7 +9,7 @@ module ApiAnalyticsTracer
       user_ip: query.context[:request_ip],
       query: query.query_string.squish,
       query_variables: query.provided_variables,
-      client_type: request.headers['dgidb-client-name']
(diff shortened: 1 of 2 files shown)


server/app/lib/api_analytics_tracer.rb
```diff
@@ -8,7 +8,8 @@ def analyze_query(query: )
     params = {
       user_ip: query.context[:request_ip],
       query: query.query_string.squish,
-      query_variables: query.provided_variables
+      query_variables: query.provided_variables,
+      client_type: request.headers['dgidb-client-name']
```
acoffman, line 12: I haven't tested it myself, but I don't believe `request` is available in here by default. You may have to grab this value in the controller and pass it in via the query context (like the ip address).

jsstevenson: this is what I get for forcing a PR out before lunch

acoffman (approved):
LGTM!

PR #538, opened 2024-11-25
fix: update api uri

jsstevenson:
@acoffman I'm guessing this is the change?


acoffman (approved):
Yep, I assume that will do the trick! Unless it's using .env.production as well? Not where why there are two there.

jsstevenson:
@katiestahl do you know why we might have both `.env.prod` and `.env.production`?

katiestahl (approved):

PR #542, opened 2024-11-26
fix: use latest GO api changes

jsstevenson:
Two weird changes seem to have happened with the GO API:

1) it doesn't like the quotes in the URL anymore
2) if pages are exhausted, it gives a 404 rather than returning an empty array

this PR adds fixes for both

### DIFF
--- a/server/lib/genome/importers/api_importers/go/importer.rb
+++ b/server/lib/genome/importers/api_importers/go/importer.rb
@@ -36,13 +36,13 @@ module Genome; module Importers; module ApiImporters; module Go;
       categories.each do |category, go_id|
         start = 0
         rows = 500
-        genes = api_client.genes_for_go_id(go_id, start, rows)
-        while genes.count.positive? do
+        loop do
+          genes = api_client.genes_for_go_id(go_id, start, rows)
           genes.each do |gene|
             create_gene_claim_for_entry(gene, category) if gene['taxon_label'] == 'Homo sapiens'
           end
+          break if genes.count < rows
           start += rows
-          genes = api_client.genes_for_go_id(go_id, start, rows)
         end
       end
     end
--- a/server/lib/genome/importers/api_importers/go/api_client.rb
+++ b/server/lib/genome/importers/api_importers/go/api_client.rb
@@ -15,7 +15,6 @@ module Genome; module Importers; module ApiImporters; module Go;
     end
 
     def make_get_request(uri)
-      puts uri
       res = Net::HTTP.get_response(uri)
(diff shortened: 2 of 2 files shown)


server/lib/genome/importers/api_importers/go/importer.rb
```diff
@@ -36,13 +36,13 @@ def create_gene_claims
...
         start = 0
         rows = 500
-        genes = api_client.genes_for_go_id(go_id, start, rows)
-        while genes.count.positive? do
+        loop do
+          genes = api_client.genes_for_go_id(go_id, start, rows)
           genes.each do |gene|
             create_gene_claim_for_entry(gene, category) if gene['taxon_label'] == 'Homo sapiens'
           end
+          break if genes.count < rows
```
acoffman, line 44: Does this break if the number of rows on the last page happens to be the same as the page size and run into the `raise StandardError` in the api client on the next request?

If so, maybe you want to handle that separately? 

You can directly check the http response code for 404 (as it is currently doing with 200), or you could match on the response type like this example [here](https://ruby-doc.org/3.3.5/stdlibs/net/Net/HTTP.html#class-Net::HTTP-label-Following+Redirection) if you find that to be cleaner. (Note it will not be redirection as the subclass but NotFound). I could be misunderstanding what's going on though.

acoffman (changes requested):
One small question, otherwise looks good!

server/lib/genome/importers/api_importers/go/importer.rb
jsstevenson: yeah, totally. I was being lazy.

it seems like the GO API returns the same response payload for a bunch of different classes of errors:

```json
{"detail":"Item with ID GO:0008233 not found"}
```

You get a variation of this ^ if you [request a nonexistent entity](https://api.geneontology.org/api/bioentity/function/GO:0008233sdflkjasdf), but also if you [make a request where "start" is beyond the number of records](https://api.geneontology.org/api/bioentity/function/GO:0008233?start=9000&rows=500). I was trying to avoid catching 404s because, like, those are nice to know about for other reasons, but maybe we can condition this such that we raise the exception if the initial request with `start=0` returns a 404, but we just break the loop once we're starting from a higher number.

jsstevenson: There's also a good chance that this is a recent bug that will get fixed, but the issue tracker for the GO site has like 50 open PRs and hundreds of open issues, so... idk

acoffman (approved):
LGTM!

PR #555, opened 2024-12-11
feat: rake task to load DB from local file

jsstevenson:
* add a rake task to load DB w/ local sql dump
* remove rake task to download from remote. It depends on a static s3://<wagner bucket>/dgidb/dgidb_latest.sql URI and I'm not really happy with maintaining that. I'd rather have a better way of dynamically pointing to the latest-dated dump. Also, I thought some of the data was supposed to be private but I could be wrong.
* slightly reconfigure the section for psql executable injection into the PATH. Drop macports support. Add homebrew (ARM64) support.


acoffman (approved):
Looks good to me, I forgot that Macports stuff was even still there!

PR #557, opened 2024-12-17
test: add basic API tests

jsstevenson:
Add simple tests for successful server response to requests made to the GraphQL API.

Slightly modified tracer configs to use a Ruby environment config rather than check directly for the environment. This value is temporarily changed in the request test but otherwise unaltered.


acoffman (approved):
Looks good to me.

Issue #567, opened 2025-01-24
Wrong xref/alias for ACELLULAR DERMAL MATRIX


vdancik:
Aliases for ACELLULAR DERMAL MATRIX contain a wrong xref, NCIT:C1326:

https://dgidb.org/drugs/ncit:C186783 (ACELLULAR DERMAL MATRIX)

NCIT:C1326 is actually DOXORUBICIN HYDROCHLORIDE/ADRIAMYCIN:

https://dgidb.org/drugs/rxcui:142433
https://ncithesaurus.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C1326 (Doxorubicin Hydrochloride)

jsstevenson:
I haven't had a chance to look at this yet. It looks a little different than the problem with the other one but the concept group it's been dumped into is pretty messy. Will circle back next week.

PR #575, opened 2025-03-06
feat: use concept IDs in grouping first

jsstevenson:
This looks like a lot of code changes but it's actually a fairly simple fix (that was made more complicated by aforementioned spaghettiness). Basically, when trying to group a drug or gene, we now do three steps

1. Collect claim name/aliases that have `nomenclature` values taken from a select set of identifier spaces (e.g., for genes, collect all name/aliases that are an HGNC ID, an NCBI ID, or an Ensembl ID). Sort them by an order of preference (for genes, HGNC > NCBI > Ensembl). Take the *first* match. This is the new part.
2. If that fails, and if the claim name wasn't one of the previously selected IDs, run it through the normalizer, and put it in that group if it matches.
3. Failing that, do the previous alias-based "voting" that we'd been falling back on previously. No change here, except this is done *only* on aliases/trade names/lesser known IDs.

### DIFF
--- a/server/lib/genome/groupers/base.rb
+++ b/server/lib/genome/groupers/base.rb
@@ -2,9 +2,32 @@ module Genome
   module Groupers
     # Provides some shared methods for groupers, shouldn't be directly instantiated.
     class Base < Genome::DataImporter
+
+      def initialize
+        @term_to_normalized_id = {}
+        @sources = {}
+      end
+
+      private
+
+      def print_grouping_message(claims, source_id)
+        if source_id.nil?
+          puts "Grouping #{claims.length} ungrouped #{@entity_type} claims"
+        else
+          begin
+            source = Source.find(source_id)
+          rescue ActiveRecord::RecordNotFound
+            puts 'Unrecognized source ID provided'
+            return
+          end
+          source_name = source.source_db_name
+          puts "Grouping #{claims.length} ungrouped #{@entity_type} claims from #{source_name}"
+        end
+      end
+
       # Sends an empty query to concept normalizer to retrieve the service's version
       def retrieve_normalizer_version
-        response = retrieve_normalizer_response('')
+        response = retrieve_normalized_concept_id('')
         response['service_meta_']['version']
       end
 
@@ -30,10 +53,37 @@ module Genome
         body['source_matches'].transform_values { |value| value['source_meta_'] }
       end
 
-      # Normalize claim terms
-      # If the primary term normalizes successfully, go with that. Otherwise, iterate
-      # through claims and tally all 'votes' for normalized concepts with the highest match
-      # type. For example:
+      # Construct individual groups for normalizing terms
+      # Returns a sorted array of normalizable concept IDs, a primary name if given,
+      # and aliases that aren't otherwise known to normalize (e.g. other names, trade names,
+      # concept IDs from niche namespaces)
+      def get_claim_search_groups(claim, claim_aliases)
+        groups = { 'concept_ids' => [], 'name' => nil, 'aliases' => [] }
+
+        concept_ids = []
+        if normalizable_ids.include? claim.nomenclature
+          concept_ids << [claim.name, claim.nomenclature]
+        else
+          groups['name'] = claim.name
+        end
+
+        claim_aliases.each do |ca|
+          if normalizable_ids.include? ca.nomenclature
+            concept_ids << [ca.alias, ca.nomenclature]
+          else
+            groups['aliases'] << ca.alias
+          end
+        end
+
+        groups['concept_ids'] = concept_ids
+                                .sort_by! { |_, nomenclature| normalizable_ids.index(nomenclature) || Float::INFINITY }
+                                .map { |concept_id| concept_id[0] }
+        groups
+      end
+
+      # Try to determine grouping of a claim based on consensus of its aliases.
+      # Iterate through aliases and tally all 'votes' for normalized concepts with the
+      # highest match type. For example:
       # 1) if a gene has three aliases, and two match as `ALIASES` to gene A and the third
       # matches as an `ALIAS` to gene B, we pick gene A
       # 2) if a gene has five aliases, and one alias matches as a `LABEL` to gene A, but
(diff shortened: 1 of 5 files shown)


server/lib/genome/groupers/base.rb
```diff
@@ -30,10 +53,37 @@ def fetch_source_meta
...
+        claim_aliases.each do |ca|
+          if normalizable_ids.include? ca.nomenclature
+            concept_ids << [ca.alias, ca.nomenclature]
+          else
+            groups['aliases'] << ca.alias
+          end
+        end
+
+        groups['concept_ids'] = concept_ids
+                                .sort_by! { |_, nomenclature| normalizable_ids.index(nomenclature) || Float::INFINITY }
```
acoffman, line 79: The `!` here will mutate the original array rather than returning a copy. It's entirely possible you intended this, it just reads as a little strange chained with the `map` which will produce a copy so I wanted to flag it in case.

server/lib/genome/groupers/drug_grouper.rb
```diff
@@ -4,53 +4,134 @@ class DrugGrouper < Genome::Groupers::Base
...
+          group_drug_claim(drug_claim)
+
+          pbar.progress += 1
+        end
+      end
+
+      private
+
+      def find_drug_attribute(drug_claim_attribute)
+        drug_attribute = DrugAttribute.where(
```
acoffman, line 34: I see that this was moved, not new code for this PR, so perhaps its not the right place to address it but aren't both branches of the `if` here checking the same thing, just expressed differently? Comparing `upper()` with `upcase` is effectively the same as comparing `lower()` with `downcase` unless there's some inconsistency with encoding somewhere.

jsstevenson:
```ruby
      def find_drug_attribute(drug_claim_attribute)
        DrugAttribute.where('name ILIKE ? AND value ILIKE ?', drug_claim_attribute.name, drug_claim_attribute.value)
                     .first
      end
```

?

acoffman (approved):
LGTM

mcannon068nw:
> This looks like a lot of code changes but it's actually a fairly simple fix (that was made more complicated by aforementioned spaghettiness). Basically, when trying to group a drug or gene, we now do three steps
> 
> 1. Collect claim name/aliases that have `nomenclature` values taken from a select set of identifier spaces (e.g., for genes, collect all name/aliases that are an HGNC ID, an NCBI ID, or an Ensembl ID). Sort them by an order of preference (for genes, HGNC > NCBI > Ensembl). Take the _first_ match. This is the new part.
> 2. If that fails, and if the claim name wasn't one of the previously selected IDs, run it through the normalizer, and put it in that group if it matches.
> 3. Failing that, do the previous alias-based "voting" that we'd been falling back on previously. Two changes here
>    
>    1. Since we're pulling out a selection of xref-like aliases above, they aren't included here
>    2. I got rid of the "tiebreaker" logic at the very bottom (where, if you have a tied # of normalizer "alias" matches, you take the one that's alphabetically earlier). When I looked at this a few months ago, it happened a total of 1 time across all drug and gene groups, but it seems like a bad idea (better to refuse to group when we're unsure).
> 
> Previously, we searched on the claim name first, but that can be an ambiguous symbol (a drug named "ADM" can resolve in multiple ways). By first searching on a selection of high-quality identifiers, we get a more reliable grouping outcome. I intend to do a little more testing, but this should close #567.

I think this makes sense to me. The tiebreaker logic at the bottom being removed feels like a good choice but should we at least make some kind of console log for this? Like it happens exactly once which is good but will it happen again in the future? (This likely doesn't matter too much just trying to think ahead).

For step 2, I think this makes sense to me. Is the idea that since it got to this step to begin with, it is clearly an erroneously attached alias and by running it through the normalizer, we are trying to find which group it actually belongs to? Would we expect this to ever 1) return the same group that the parent would match anyway or 2) return no group at all

jsstevenson:
> should we at least make some kind of console log for this?

Yeah I was thinking the same thing, I added some logging to this section so we can be generally more aware of when stuff is falling through the cracks.

jsstevenson:
[changed_drug_groupings.csv](https://github.com/user-attachments/files/19253333/changed_drug_groupings.csv)
[changed_gene_groupings.csv](https://github.com/user-attachments/files/19253334/changed_gene_groupings.csv)

these show instances where the group ID changes as a result of this PR. Few for gene, but lots for drug. From sampling them, it looks like it's usually a case where the normalizer hasn't successfully merged groups of IDs that otherwise should be merged, and the PR results in simply changing which of the incompletely-merged groups the claim gets put into it. Which is to say, I am not finding instances where the PR results in a 'wrong' choice yet, but it is a change.

mcannon068nw:
From looking through the drug list, it looks like the majority change is for groupings is from rxcui to chembl identifiers:

`{'drugbank': -206, 'drugsatfda.nda': -1, 'iuphar.ligand': 154, 'ncit': -253, 'wikidata': -18, 'rxcui': -3747, 'hemonc': 0, 'drugsatfda.anda': 2, 'chembl': 4116, 'chemidplus': -2}`

From doing some spot checking and manual investigation of these rxcui -> chembl changes, it seems like this is likely addressing failures to join groups and/or formulation differences (e.g. hydrochloride, bromide, maleate etc.)
Screenshot 2025-03-17 at 9 33 02 AM

An example of the old normalized group for Fluoxitine Hydrochloride:
OLD: rxcui:227224
```
{
  "query": "rxcui:227224",
  "warnings": [],
  "match_type": 100,
  "service_meta_": {
    "name": "thera-py",
    "version": "0.9.0",
    "response_datetime": "2025-03-17T13:34:36.020185Z",
    "url": "https://github.com/cancervariants/therapy-normalization"
  },
  "therapy": {
    "id": "normalize.therapy.rxcui:227224",
    "extensions": [
      {
        "name": "aliases",
        "value": [
          "FLUoxetine",
          "Fluoxétine",
          "APRD00530",
          "NSC-283480",
          "NSC-758685",
          "FLUOXETINE",
          "Fluoxetine",
          "Fluoxetina",
          "Fluoxetine hydrochloride",
          "Fluoxetine Hydrochloride",
          "Fluoxetinum",
          "FLUoxetine hydrochloride",
          "Fluoxetin ratiopharm",
          "fluoxetine",
          "Fluoxetin",
          "Sarafem",
          "Symbyax",
          "FLUOXETINE HYDROCHLORIDE",
          "Fluval",
          "N-methyl-gamma-(4-(trifluoromethyl) -phenoxy)-benzenepropanamine",
          "(+-)-N-Methyl-3-phenyl-3-((alpha,alpha,alpha-trifluoro-P-tolyl)oxy)propylamine",
          "(+-)-N-Methyl-gamma-(4-(trifluoromethyl)phenoxy)benzenepropanamine",
          "N-methyl-3-phenyl-3-[4-(trifluoromethyl)phenoxy]propan-1-amine",
          "Prozac"
        ]

```
NEW: chembl:CHEMBL1201082
```
{
  "query": "chembl:CHEMBL1201082",
  "warnings": [],
  "match_type": 100,
  "service_meta_": {
    "name": "thera-py",
    "version": "0.9.0",
    "response_datetime": "2025-03-17T13:35:37.319084Z",
    "url": "https://github.com/cancervariants/therapy-normalization"
  },
  "therapy": {
    "id": "normalize.therapy.chembl:CHEMBL1201082",
    "extensions": [
      {
        "name": "aliases",
        "value": [
          "Adofen",
          "Fluctin",
          "Fluneurin",
          "Fluox-puren",
          "Fluoxeren",
          "Fluoxetine (as hydrochloride)",
          "Fluoxetine hydrochloride",
          "Flusol",
          "Fluxet",
          "Fontex",
          "Foxetin",
          "LY-110140",
          "LY110140",
          "Lovan",
          "Reconcile"
        ]
      },
      {

```

Given the above, I think this change is a net positive and will result in the joining of a fair amount of fragmented groups.

PR #577, opened 2025-03-14
Exclude unused CIViC Therapies

### LINKED ISSUE #576: Don't import CIViC therapies that aren't included in evidence/assertions
CIViC includes a bunch of therapy entries for typeahead/future curating purposes that aren't yet included in any evidence. Because they don't add any additional metadata to those terms, there's no reason for us to import them. We should update the importer logic to exclude these terms.

acoffman:
This utilizes a new filter added in griffithlab/civic-v2#1212 to exclude Therapies from import unless they have at least one linked, non-rejected Evidence Item or Assertion.

We won't be able to deploy it until the referenced PR is merged and deployed in CIViC; I will follow up here when it is.

closes #576 

(Test failure is the result of the filter not existing on the prod CIViC schema yet)

### DIFF
--- a/server/lib/genome/importers/api_importers/civic/api_client.rb
+++ b/server/lib/genome/importers/api_importers/civic/api_client.rb
@@ -52,7 +52,7 @@ module Genome; module Importers; module ApiImporters; module Civic
 
     DrugsQuery = CivicApi::Client.parse <<-GRAPHQL
       query ($after: String!) {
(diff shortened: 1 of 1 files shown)


acoffman:
The corresponding CIViC PR has been deployed, and now the tests pass here as the schema matches what's expected. Should be ready for review.

jsstevenson (approved):

PR #592, opened 2025-03-31
fix: Lazy load CIViC GraphQL queries

acoffman:
The ruby graphql client library generally recommends parsing queries against the schema at boot to avoid runtime errors and ensure you're not accepting arbitrary user injected strings (https://github.com/github-community-projects/graphql-client/blob/master/guides/dynamic-query-error.md). 

This unfortunately adds a boot-time dependency on connecting to the CIViC instance to retrieve the schema. The recommended workaround in the README is to include a static JSON dump of the schema in the app: https://github.com/github-community-projects/graphql-client?tab=readme-ov-file#configuration. This seems like it would be a pain point, having to keep this file up-to-date every time we make a change to the CIViC schema.

Given that these are not user-input queries but internally created for the purpose of the importer, I think its probably an okay compromise in this case to turn the "allow dynamic queries" flag on and lazily load these to eliminate the dependency on CIViC unless you're actively running the CIViC importer.

This PR removes the static queries, client, and schema objects and instead initializes them only when a new instance of the CIViC ApiClient class is created.

### DIFF
--- a/server/lib/genome/importers/api_importers/civic/api_client.rb
+++ b/server/lib/genome/importers/api_importers/civic/api_client.rb
@@ -3,135 +3,144 @@ require "graphql/client/http"
 
 module Genome; module Importers; module ApiImporters; module Civic
   class ApiClient
+    attr_reader :genes_query, :drugs_query, :interactions_query, :client
+
+    def initialize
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):
Great! 👍

PR #594, opened 2025-04-10
fix: use correct civic source type enum

### LINKED ISSUE #588: Increase interaction claim publication annotations
We only ingest publications for a small number of sources. It's not possible to get PMIDs for every source, but there are some easy lifts we could be doing (for example, the publication-based sources should all have a static PMID for every interaction).

Making this in re an email query from a user

jsstevenson:
close #588 

We were previously checking that the source type was `"PubMed"` but it appears to be `"PUBMED"` now. I'm wondering if we could code this into the graphql query in a way that references the enum rather than hardcoding it on our side.

Also made some small style changes that are purely a distraction.

### DIFF
--- a/server/lib/genome/importers/api_importers/civic/importer.rb
+++ b/server/lib/genome/importers/api_importers/civic/importer.rb
@@ -60,7 +60,7 @@ module Genome
             if ev_item.source.citation_id.present? && ev_item.source.source_type == 'PUBMED'
               create_interaction_claim_publication(ic, ev_item.source.citation_id)
             end
-            create_interaction_claim_link(ic, "ev_itemD#{ev_item.id}", "https://civicdb.org/evidence/#{ev_item.id}")
+            create_interaction_claim_link(ic, "EID#{ev_item.id}", "https://civicdb.org/evidence/#{ev_item.id}")
           end
 
           # current policy is to skip items with poor rating/level, anything that doesn't support the claim,


server/lib/genome/importers/api_importers/civic/importer.rb
```diff
@@ -55,39 +55,40 @@ def create_gene_claims
...
-            ic = create_interaction_claim(gc, dc)
-            if ei.source.citation_id.present? && ei.source.source_type == 'PubMed'
-              create_interaction_claim_publication(ic, ei.source.citation_id)
+          def create_entries_for_evidence_item(ev_item, drug_claim, gene_claim)
+            ic = create_interaction_claim(gene_claim, drug_claim)
+            if ev_item.source.citation_id.present? && ev_item.source.source_type == 'PUBMED'
+              create_interaction_claim_publication(ic, ev_item.source.citation_id)
             end
-            create_interaction_claim_link(ic, "EID#{ei.id}", "https://civicdb.org/evidence/#{ei.id}")
+            create_interaction_claim_link(ic, "ev_itemD#{ev_item.id}", "https://civicdb.org/evidence/#{ev_item.id}")
```
acoffman, line 63: Is the capital `D` supposed to remain here after `ev_item`?

acoffman (approved):
Looks good to me assuming the one mentioned item is in fact intentional. We can also add a source type filter to the API as discussed in slack if you would prefer!

server/lib/genome/importers/api_importers/civic/importer.rb
jsstevenson: not intentional!

PR #598, opened 2025-04-23
feat: add stats api endpoint

### LINKED ISSUE #597: Add item stats endpoint
* Probably just put it in as a new rails controller rather than graphql for simplicity
* use rails caching https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching-using-rails-cache
* could get away with a lot vis-a-vis caching since the data rarely changes

include the following

```
Drug Claims
SELECT COUNT(*) FROM drug_claims

Drug Groups
SELECT COUNT(*) FROM drugs

Gene Claims
SELECT COUNT(*) FROM gene_claims

Gene Groups
SELECT COUNT(*) FROM genes

Interaction Claims
SELECT COUNT(*) FROM interaction_claims

Interaction Groups
SELECT COUNT(*) FROM interactions

Gene Category Claims
SELECT COUNT(*) FROM gene_claim_categories_gene_claims

Gene Category Groups
SELECT COUNT(*) FROM gene_categories_genes

Sources
SELECT COUNT(*) FROM sources

PMIDs
SELECT COUNT(*) FROM publications
```

jsstevenson:
close #597 

it's cheap to compute so I figure invalidating the cache daily is fine

@mcannon068nw let me know if you want any other things tallied here

### DIFF
--- a/server/app/controllers/counts_controller.rb
+++ b/server/app/controllers/counts_controller.rb
@@ -1,0 +1,20 @@
+class CountsController < ApplicationController
+  def index
+    counts = Rails.cache.fetch('api_counts', expires_in: 24.hours) do
+      {
+        drug_claims: DrugClaim.count,
+        drugs: Drug.count,
+        gene_claims: GeneClaim.count,
+        genes: Gene.count,
+        interaction_claims: InteractionClaim.count,
+        interactions: Interaction.count,
+        gene_categorization_claims: GeneClaimCategory.joins(:gene_claims).count,
+        gene_categorizations: GeneClaimCategory.joins(:genes).count,
+        sources: Source.count,
+        publications: Publication.count
+      }
+    end
+
+    render json: counts
+  end
+end
--- a/server/config/routes.rb
+++ b/server/config/routes.rb
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
-  post "/api/graphql", to: "graphql#execute"
-  post "/api/graphql/download/:table_name", to: "table_download#download_table"
+  post '/api/graphql', to: 'graphql#execute'
+  post '/api/graphql/download/:table_name', to: 'table_download#download_table'
 
-  mount GraphiQL::Rails::Engine, at: "/api/graphiql", graphql_path: "/api/graphql"
+  get '/api/counts', to: 'counts#index'
+
+  mount GraphiQL::Rails::Engine, at: '/api/graphiql', graphql_path: '/api/graphql'
 end


server/app/controllers/counts_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def index
+    counts = Rails.cache.fetch('api_counts', expires_in: 24.hours) do
+      {
+        drug_claims: DrugClaim.count,
+        drugs: Drug.count,
+        gene_claims: GeneClaim.count,
+        genes: Gene.count,
+        interaction_claims: InteractionClaim.count,
+        interactions: Interaction.count,
+        gene_categorization_claims: GeneClaimCategory.joins(:gene_claims).count,
```
acoffman, line 11: For this, and the query below, are you trying to count the number of unique categories that are linked to a gene claim (or gene) or are you trying to count the total number of linkages?

Basically, 

```ruby
GeneClaimCategory.joins(:gene_claims).count
```
 will generate: 

```sql
SELECT COUNT(*) FROM "gene_claim_categories" INNER JOIN "gene_claim_categories_gene_claims" ON "gene_claim_categories_gene_claims"."gene_claim_category_id" = "gene_claim_categories"."id" INNER JOIN "gene_claims" ON "gene_claims"."id" = "gene_claim_categories_gene_claims"."gene_claim_id"`
```

which will count every single link between a gene claim and a gene claim category vs 

```ruby
 GeneClaimCategory.joins(:gene_claims).distinct.count
```

which will generate

```sql
SELECT COUNT(DISTINCT "gene_claim_categories"."id") FROM "gene_claim_categories" INNER JOIN "gene_claim_categories_gene_claims" ON "gene_claim_categories_gene_claims"."gene_claim_category_id" = "gene_claim_categories"."id" INNER JOIN "gene_claims" ON "gene_claims"."id" = "gene_claim_categories_gene_claims"."gene_claim_id"
```

and count how many distinct categories are linked to at least one gene claim. Both seem like totally valid counts to display, just wanted to make sure we were displaying the one we intend to.

acoffman (approved):
Looks good to me!

server/app/controllers/counts_controller.rb
mcannon068nw: The initial intent was to have the number of claims attached to at least one gene category but upon review yeah that initial SQL statement wouldn't get us that. I like the change to distinct categories with at least one gene claim tbh

mcannon068nw (approved):

server/app/controllers/counts_controller.rb
jsstevenson: Yeah, we went back and forth a little bit on this, we'll probably change our minds again anyway

PR #608, opened 2025-08-27
Add survey alert to homepage

acoffman:
After some discussion in slack, this is the layout I went with:

Screenshot 2025-08-27 at 9 20 13 AM


client/src/components/Shared/SurveyAltert/SurveryAlert.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const SurveyAlert: React.FC = () => {
+  return (
+    <div className="survey-alert">
+      <Alert
+        severity="info"
+        icon={
+          <img
+            src="/images/survey.png"
+            alt="Survey Logo"
+            style={{ width: 93, height: 71 }} //this is just 50% scale for the image, preserving aspect ratio
```
katie-perry: I think you could also use width: 50%, height: auto to accomplish this

katie-perry (approved):
looks awesome! Tested on a variety of screen sizes, and the alert is nice and responsive and looks solid. Nice work :)

client/src/components/Shared/SurveyAltert/SurveryAlert.tsx
acoffman: Ah yep, that's much better than just having a magic number hardcoded in there. Thanks!

mcannon068nw (approved):
Looks great to me!

PR #617, opened 2026-01-12
feat: add date to latest row on download page

### LINKED ISSUE #612: Add indication of date to "latest" download entry on downloads page
Currently it could be ambiguous whether latest is a duplicate of 2024-Dec row. We should add some indication of date for the latest row.

katie-perry:
This is to close #612 

This is a temporary solution. There's not a great solution currently to be able to actually get the date for the latest data, so we're going to hardcode it for now. We should create a followup ticket to be able to get the date from the file some other way.


mcannon068nw (approved):

katie-perry:
followup ticket is #618

PR #628, opened 2026-04-29
Update to current stable ruby and rails

acoffman:
This brings the backend application up-to-date from the unsupported rails 6 and ruby 3 and gets it on the latest stable version of each.

Most of the changes are minimal and revolve around changes to how autoloading and inflection have changed over the years. The "large" diff is to the Pharos importer which has always been a little complicated. The GraphQL Client library wants things defined as constants, but that means it actually makes a network call when the file is loaded/parsed. This change works around that issue.

Tests are passing locally, the app boots, and browsing around the UI, all the major views still seem to function. Last bit I probably need to change are the installation instructions which are now out of date.


jsstevenson (approved):
👍 sure!

acoffman:
Does anyone have a tooling preference for the README instructions? `rbenv` is more popular than `rvm` at this point and doesn't require as many shell antics. Personally, I use `mise` for this sort of thing nowadays, but am also happy to just stick with `rvm` for the README if that's what folks are using at Nationwide

jsstevenson:
> Does anyone have a tooling preference for the README instructions? rbenv is more popular than rvm at this point and doesn't require as many shell antics. Personally, I use mise for this sort of thing nowadays, but am also happy to just stick with rvm for the README if that's what folks are using at Nationwide

I'm sure we can work with whatever you think is best -- I got frustrated at one point and switched over to `asdf` but I'm not exactly in love with it

acoffman:
I updated the README with instructions for using `mise` if someone wants to take a look. In theory it should now manage ruby, node, python, and the virtualenv automatically. Don't want to force folks to use it though so feel free to object and I'll change it back.

I did have an issue running through the instructions trying to install the requirements.txt:

```
ERROR: Could not find a version that satisfies the requirement psycopg-binary==3.2.3 (from versions: 3.2.10, 3.2.11, 3.2.12, 3.2.13, 3.3.0, 3.3.1, 3.3.2, 3.3.3)
```

Not sure if I pinned the wrong version of python or if there's something else going on there.

PR #640, opened 2026-08-24
Add Docker configuration

acoffman:
This PR adds dockerfiles and supporting configuration/scripts to build the containers. There are two containers: one contains the application itself and the other contains postgres and the data dump baked into it already. There is a docker compose config that should allow a user to spin up a local instance with a single command. There are some minor changes to the server to accomodate this since we are not using docker in production and the server/assets sit behind a reverse proxy there.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+.git
+.github
+.venv
+**/.DS_Store
+
+client/build
+client/node_modules
+client/.yarn/install-state.gz
+
+server/.bundle
+server/log/*
+server/public/*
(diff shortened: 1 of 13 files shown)


mcannon068nw (approved):
Nice, followed the README.md and it all works nicely. Searches work as well on the localhost

Image

Image

PR #648, opened 2026-09-01
feat: add initial prism source

jsstevenson:
* Using table 4 from the supplementary data in https://pmc.ncbi.nlm.nih.gov/articles/PMC7328899/#SM1, create a TSV
* Ingest all interactions from feature sets `PROT, GE, CN, DEMETER2_COM, AVANA_PUBLIC_18Q2, MUTPOOL, MUT, MUTDMG, MUTHOT, MUTMIS, METHYL` with pearson score > 0.2 (cutoff mentioned directly in the paper)
* Include interaction claim attributes describing assay metadata and original feature description

there is a little work that could be done to integrate table 5 annotations; I think it can wait though

### DIFF
--- a/server/lib/genome/importers/file_importers/prism.rb
+++ b/server/lib/genome/importers/file_importers/prism.rb
@@ -40,7 +40,11 @@ module Genome
             'CN' => 'CN',
             'DEMETER2_COM' => 'DEM2',
             'AVANA_PUBLIC_18Q2' => 'CRISPR',
-            'MUTPOOL' => 'MUTpool'
+            'MUTPOOL' => 'MUTpool',
+            'MUT' => 'RMUTmis',
+            'MUTMIS' => 'RMUTmis',
+            'MUTHOT' => 'RMUThot',
+            'MUTDMG' => 'RMUTdmg'
           }.freeze
 
           def ingest_ncbi_feature(feature_name, prefix)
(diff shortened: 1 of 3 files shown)


server/lib/genome/importers/file_importers/prism.rb
```diff
@@ -0,0 +1,202 @@
...
+
+            gene_claim
+          end
+
+          # Feature names include a bunch of extra window dressing;
+          # We need to parse out the actual genomic feature differently depending on feature set type
+          def ingest_feature(feature_name, feature_set)
+            normalized_feature_set = feature_set.strip.upcase
+            case normalized_feature_set
+            when 'GE'
```
mcannon068nw, line 71: I love a good case switch!

server/lib/genome/importers/file_importers/prism.rb
```diff
@@ -0,0 +1,202 @@
...
+              )
+              create_interaction_claim_attribute(interaction_claim, InteractionAttributeName::CONTEXT,
+                                                 context_value)
+
+              assay_value = assay_details_value(
+                pearson_score: row['pearson_score'].strip,
+                sample_size: row['sample_size'],
+                screen_id: row['screen_id'],
+                dose: row['dose']
+              )
```
mcannon068nw, line 193: I would wager these are always present values, but is there any protection against missing values here just in case? or logging in case this happens for some reason

server/lib/genome/importers/file_importers/prism.rb
```diff
@@ -0,0 +1,202 @@
...
+          # eg "ATLANTIS predictive model; Pearson score: 0.42; sample size: 123; screen: MTS010; dose: 2.5 µM"
+          def assay_details_value(pearson_score:, sample_size: nil, screen_id: nil, dose: nil)
+            details = [
+              'ATLANTIS predictive model',
+              "Pearson score: #{pearson_score}"
+            ]
+
+            details << "sample size: #{sample_size}" if sample_size.to_s.strip != ''
+            details << "screen: #{screen_id}" if screen_id.to_s.strip != ''
+            details << "dose: #{dose}" if dose.to_s.strip != ''
```
mcannon068nw, line 163: Do we want to retain the BROAD_ID here? (e.g. BRD-A00055058-001-01-0).

This might mean something to someone or provide additional provenance on the exact source

mcannon068nw (changes requested):
Want your thoughts on two questions in the comments

server/lib/genome/importers/file_importers/prism.rb
jsstevenson: yeah the `assay_details_value` function guards against cases where any of these is an empty string

server/lib/genome/importers/file_importers/prism.rb
jsstevenson: Yeah, let's add this when we incorporate the stuff from the table join that Anastasia will be working on

(We could also add the same on the Drug Repurposing Hub side -- their "sample information" file includes this + a pubchem id which would be nice https://repo-hub.broadinstitute.org/repurposing#download-data)

mcannon068nw (approved):

Issue #644, opened 2026-08-24
Reference the DGIdb MCP Paper


mcannon068nw:
Also add this into the survey banner? 

Make an announcement of the MCP paper

jsstevenson:
It would be really nice to just have some "news" cards

Issue #584, opened 2025-03-26
Migrate links to top bar


mcannon068nw:
As elements shift around the page and the carousel is introduced, we should migrate links to the topbar per: https://www.figma.com/design/GutMA6FI3PhJNEmZM1FXiV/DGIdb-Wireframes?node-id=151-71&t=PWKYDabMMBCEWw8L-0

Additionally, we need to display the new page divisions for the About in some way, potentially the About being a drop down? Open to ideas for this.

PR #655, opened 2026-09-10
feat: add stats and news feed to homepage, clean up width/spacing

jsstevenson:
close #584
close #644 

Made some changes based on some wireframe work done by @katie-perry a while ago (https://www.figma.com/design/GutMA6FI3PhJNEmZM1FXiV/DGIdb-Wireframes?node-id=147-2)

* Fix some spacing/padding/width stuff for cleanliness
* Remove GitHub and API links from homepage; add API link to header bar
* Add item quantities ticker
* Add rudimentary news feed (see below)

### News feed

Very rudimentary, intended for 2-3 line blurbs about new stuff. Best used for announcements, but not currently a good fit for "blog" style posts that go into depth about features or use cases. I added two items to start: an announcement of the MCP paper, and an announcement for the 5.0.13 release (prospective).

I thought about a lot of different approaches to this, but I settled on the most naive one to keep things simple, and because I don't think the expected output justifies a more complicated workflow. There is an array, `newsItems`, in the `NewsFeed.tsx` module that declares objects which define individual feed items. Because everything lives in source code, any news about a new release needs to be committed and merged as part of that release.

Only the 3 newest news feed items are displayed by default; there's a button to show/hide the rest if the length of the array is > 3.

Screenshot 2026-09-10 at 3 30 52 PM


jsstevenson:
I think this is *not* intended to merge yet but a little review won't hurt

